PR #4, opened 2018-04-10
Convert default bindings to use scancode-based keys.

slipher:


default.cfg
```diff
@@ -30,32 +30,30 @@ bind  CTRL          +movedown
...
+bind  hw:q          +activate
+bind  hw:e          "modcase alt \"deconstruct marked\" deconstruct"
+bind  hw:c          +taunt
 
 //
 // CLIENT ENVIRONMENT COMMANDS
 //
 
 bind  PAUSE         pause
-bind  ESCAPE        togglemenu
```
illwieckz: is togglemenu keybind not needed anymore?

slipher: Binding Escape has no effect, it is ignored (this was true before my changes as well).

PR #7, opened 2019-06-23
Press escape to cancel connecting to a server

slipher:
The implementation in the engine [1] stopped working since the ui key
catcher is now enabled during connection.

[1] https://github.com/DaemonEngine/Daemon/blob/b17dd395a247ad0e0a53435d13ce73efb0863cb8/src/engine/client/cl_keys.cpp#L602

### DIFF
--- a/ui/transition_connecting.rml
+++ b/ui/transition_connecting.rml
@@ -22,7 +22,9 @@
 			}
 		</style>
 	</head>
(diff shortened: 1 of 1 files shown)


illwieckz:
oh thank you that was annoying to not be able to cancel this!

PR #8, opened 2019-11-09
display price in evolve menu

Gireen:
solves partly  Unvanquished/Unvanquished#1123
but displays a price of 0 for locked classes


illwieckz:
I noticed it prints uselesses `.0 ` suffixes, but that's probably not an asset issue

Gireen:
yeah. it was meant to look like the morph points displayed in the hud

PR #12, opened 2020-02-04
Remove refresh button

Gireen:
fix unvanquished/unvanquished#1156

### DIFF
--- a/ui/server_browser.rml
+++ b/ui/server_browser.rml
@@ -41,9 +41,8 @@
 				</datagrid>
 
 				<buttonbar>
-					<button onClick='Events.pushevent("buildDS server_browser internet", event)'>Refresh list</button>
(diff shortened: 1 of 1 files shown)


illwieckz:
Maybe remove the refresh button in LAN browser too? Also, I now think that once the "Refresh" button is removed, the remaining all-in-one button can just be named "Refresh" (and there is no need in the button to say it's a server list)

To be honest, my dream would be to have only one list, merging LAN and Internet, like this (Quetoo):

[Quetoo server list](https://dl.illwieckz.net/b/quetoo/screenshots/20171227-121016-000.quetoo-server-browser.png)

Having only one all-in-one "Refresh" button (for Internet and LAN) would be already a step forward in that direction: Before merging those lists, we can already start to implement the behavior of having one "Refresh" button that refreshes both list, and make sure that when people click on the "Server listings" link on the left menu, local server lookup is done at the same time internet server lookup is done.

PR #14, opened 2020-04-13
ask for nickname on first start

Gireen:
set a more unique name if none is chosen
for  Unvanquished/Unvanquished#1111

this would also be a good place to state that this game is still an alpha or something similar.

### DIFF
--- a/ui/lua/util.lua
+++ b/ui/lua/util.lua
@@ -43,3 +43,13 @@ function detectEscape(event, document)
 		end
 	end
 end
+
+function welcome(event, document)
+  if Cvar.get("cg_welcome") ~= "1" then
+    if Cvar.get("name") == "UnnamedPlayer" then
+      Cvar.set("name", "Player#" .. math.ceil(math.random()*10000000))
+    end
+    Events.pushevent("show options_welcome", event)
+    Cvar.set("cg_welcome", "1")
+  end
+end
--- a/ui/menu_main.rml
+++ b/ui/menu_main.rml
@@ -5,7 +5,7 @@
 		<link type="text/rcss" href="menu.rcss" />
 	</head>
 
-	<body id="menu_main" class="mainmenu" nohide onShow='Events.pushevent("play 2 sound/ui/heartbeat sound/ui/heartbeat1", event)'>
+	<body id="menu_main" class="mainmenu" nohide onShow='welcome(event, document) Events.pushevent("play 2 sound/ui/heartbeat sound/ui/heartbeat1", event)'>
 		<img class="gradient" src="/ui/assets/background/bg" />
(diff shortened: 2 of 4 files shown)


illwieckz:
Great! That's very precious!

What's weird is that it first displays `UnnamedPlayer`:

[nickname at start](https://dl.illwieckz.net/b/unvanquished/shots/nickname-at-start/unvanquished_2020-04-13_134647_000.jpg)

Then if I click on the form, the random nick appears:

[nickname at start](https://dl.illwieckz.net/b/unvanquished/shots/nickname-at-start/unvanquished_2020-04-13_134659_000.jpg)

At least it works. I tested how it behaves if I directly load a map from command line (that would be similar to joining a server from a third-party server browser), it does not asks for the name (that's ok) but at least it sets the random nickname.

The next level would be to generate some nickname like GitHub does with project name suggestion:

[name proposal](https://dl.illwieckz.net/b/unvanquished/shots/nickname-at-start/20200413-135406-000.github-project-name-suggestion.png)

But that would require more than just UI stuff. What you done is already good for next release!

May be modify punctuation this way?

```diff
- Hello there.
+ Hello there!
- How should we call you.
+ How should we call you?
```

> This would also be a good place to state that this game is still an alpha or something similar.

Yes, but I think it may be better to permanently mark it on the background (like bottom left, with a colorful bold font, like red: “This game is in public beta”). We can even customize it: “This game is in public beta, something something <nickname>”, like if it was very special and player was so lucky to be a beta tester and how much he must be proud of it. :grin:

illwieckz:
About “Hello there!”, we may also investigate some sentences that may be more punchy, like “Get ready!”.

Or do it like if player was already in game, with some combat-related wording: “Enlist now!”. A bit like I did [on home page](https://unvanquished.net/).

What would be absolutely cool would be to get the server browser popping up right after the player has registered its name. :smiley:

Gireen:
I think that would not match for just a question about the name.  That would be more something for a dialog that enplanes how the game works.

That the server browser opens at first is probably something for another place but if its about the first interaction we could add a big button on the bottom of this welcome message that opens direct the server browser.

illwieckz:
> but if its about the first interaction we could add a big button on the bottom of this welcome message that opens direct the server browser.

Yes!

illwieckz:
@Gireen, can you add the button to open the server list? The button label may be something like “Join a server”.

Maybe add a title before, like “Let's play!” so there is a clear distinction with the previous sentence (about options).

illwieckz:
This is good enough, it can be improved later.

illwieckz:
Hmm, I forgot to change target branch name to `0.52.0/sync`, sorry for branch merge commit noise…

PR #19, opened 2020-05-04
ui: add a message saying the game is in development

illwieckz:
Add a message saying the game is in early access, idea from a comment in #14.

We were afraid to make the game much known because people may not understand the game is not finished and may build bad reputation that will stick based on wrong assumptions or things that we already know ourselves they must be fixed. But time has passed and players are now more accustomed to the “early access” thing that is now well established [see wikipedia article](https://en.wikipedia.org/wiki/Early_access).

The idea is to display a disclaimer on the bottom left of the game screen so it would be obvious for anyone that the game they play is not a complete release. This way we can be a bit braver to advertise the game.

Also, because the human people is a weird kind of social animal with irrational behaviour, it may also increases the proud of owning this game, the need to share this proud and attract interest in it. :sweat_smile:

The message is displayed on the bottom left of the screen when the menu is displayed (not when playing game of course):

[disclaimer](https://dl.illwieckz.net/b/unvanquished/wip/disclaimer/unvanquished_2020-05-04_023350_000.jpg)

[disclaimer](https://dl.illwieckz.net/b/unvanquished/wip/disclaimer/unvanquished_2020-05-04_023440_000.jpg)

### DIFF
--- a/ui/menu_ingame.rml
+++ b/ui/menu_ingame.rml
@@ -12,6 +12,8 @@
 		<img class="circles" src="/ui/assets/background/circles2" />
 		<img class="black2" src="/ui/assets/background/black2" />
 
+		<div style="position: absolute; bottom: 1em; left: 1em; color: rgba(255, 0, 0, 100); font-size: 80%;">This game version is part of the Unvanquished beta development program, game experience and depicted assets may not represent the final game.</div>
+
 		<sidebar>
 				<!-- HACK: Using shared/window.rml as a stylesheet rather than a template -->
 				<window id="innersidebar" class="innersidebar" >
--- a/ui/menu_main.rml
+++ b/ui/menu_main.rml
@@ -16,6 +16,8 @@
 		<img class="black2" src="/ui/assets/background/black2" />
 		<img class="logo" src="/ui/assets/logos/fuzzy_blue" />
 
+		<div style="position: absolute; bottom: 1em; left: 1em; color: rgba(255, 0, 0, 100); font-size: 80%;">This game version is part of the Unvanquished beta development program, game experience and depicted assets may not represent the final game.</div>
+
 		<sidebar>
 				<!-- HACK: Add some extra spacing atop the main menu -->
 				<hackyspacer />


illwieckz:
Please make comment about the wording!

slipher:
[Wiki says](https://en.wikipedia.org/wiki/Early_access):
> Early access, also known as early funding, alpha-access, alpha founding, or paid-alpha, is a funding model in the video game industry by which consumers can purchase and play a game in the various pre-release development cycles, such as pre-alpha, alpha, and/or beta, while the developer is able to use those funds to continue further development on the game.

So this term is not accurate. We should not larp as a for-profit game :stuck_out_tongue: I wouldn't say anything about a "final version" either, since usually development just trails off rather than being declared "done". Just say that we are in alpha, under development or whatever.

illwieckz:
> [Wiki says](https://en.wikipedia.org/wiki/Early_access):
> > Early access, also known as early funding, alpha-access, alpha founding, or paid-alpha, is a funding model in the video game industry by which consumers can purchase and play a game in the various pre-release development cycles, such as pre-alpha, alpha, and/or beta, while the developer is able to use those funds to continue further development on the game.
> 
>  So this term is not accurate. We should not larp as a for-profit game :stuck_out_tongue: I wouldn't say anything about a "final version" either, since usually development just trails off rather than being declared "done". Just say that we are in alpha, under development or whatever.

I'm not happy with “final version” too.

For “early access”, in fact I had to read that Wikipedia page to discover that the page makes such implication. I don't think we have to leave to others the monopoly of such expression, and I don't think that's fair. We are clearly in testing time and the release is to get players put their hand on it for real. I'm OK for other words but I chose “early access” because I think “alpha version” is not explicit enough, I want an, alternative to “early access” to be more explicit than “alpha”.

There is free-to-play games on [Steam's «Early Access» category](https://store.steampowered.com/genre/Early%20Access/), I'll add notices on the wikipedia talk page about the fact the page seems to make some wrong assumption about the funding-only nature of it.

illwieckz:
I fixed Wikipedia page, see introduction and steam early access chapter. Steam's Early Access category [features plenty of free and free-to-play games](https://store.steampowered.com/search/?sort_by=Released_DESC&term=Free&genre=Early+Access).

slipher:
I looked at some of those Steam ones and I think most of theme are for-profit developers who are allowing the game to be played for free only during the testing period. But I'll grant that some small number of truly free games with "early exist"; however, this does not affect my position. It's about how the term is commonly understood. Quoting from the top two definitions on [Urban Dictionary](https://www.urbandictionary.com/define.php?term=early%20access): 
> When video game developers sell a game before it is done to milk more money out of their customers.

> A term that is frequently used by lazy and greedy game developers that has two major benefits. They can respond to any complaints regarding issues with the game itself with "it's in early access, of course you'll approach bugs" or release the game in early access to make some quick money

And the term early access implies that the current access is somehow on different terms than later access. This isn't really true for us. As free software, anyone can access it at any time.

illwieckz:
Basically this kind of distinction is just another way to use language in a way free software looks like second class software. I'm more concerned about the “greedy”, “lazy” and “milk” wording, to be honest, as I don't want our game to be associated with that. That is an argument.

illwieckz:
[disclaimer](https://dl.illwieckz.net/b/unvanquished/wip/disclaimer/unvanquished_2020-05-04_055200_000.jpg)

String is now:

> This version is part of the Unvanquished beta development program, game experience and depicted assets may not represent the final game.

I'm not very happy with “final game” but that's the best I can find, and well, that does not mean there will no updates after 1.0 release. And I prefer “final game” than “1.0 release” that means nothing. I truly hope everyone understands that today. People even forgotten that in the 90's the cartridges were not updatable after being bought, or people were not born yet to have experienced it.

We talk there about the product to be shipped when development is considered to be complete for the milestone we would have set for completion. Even “complete game” feels weird, especially since it means to be extended by anyone, with mapping tools being easily available etc.

PR #25, opened 2020-09-10
Fix some recent additions to the menus

slipher:
See also https://github.com/Unvanquished/Unvanquished/pull/1213.

### DIFF
--- a/ui/lua/util.lua
+++ b/ui/lua/util.lua
@@ -45,11 +45,10 @@ function detectEscape(event, document)
 end
 
 function welcome(event, document)
-  if Cvar.get("cg_welcome") ~= "1" then
-    if Cvar.get("name") == "UnnamedPlayer" then
-      Cvar.set("name", "Player#" .. math.ceil(math.random()*10000000))
-    end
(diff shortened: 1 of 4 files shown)


illwieckz:
I'm still looking for a wording that tells the current status has not yet reached the finishing level a player can expect from a non-beta game. I want this to be explicit and not only implicit from the fact it's alpha/beta.

I don't know how much “finishing level” is about “there is nothing after that” in English. In french, le “niveau de finition” is a bit like a “polishing level”, that does not mean it's dead.

slipher:
"Finish" indeed means to reach the end of something, implying there is nothing after.

illwieckz:
> @slipher:
> illwieckz, is it ok to merge?

We need a better wording where the message about the experience not being on par with what we want to provide is not implied.
The fact the game is in beta state implies this is not perfect, I would like a wording being more explicit about “it's not perfect yet but we want it better so be kind”…

slipher:
Merging now since there are some important bugfixes, while the beta wording is not so critical and can be easily changed later.

PR #26, opened 2020-09-14
let alien bots defend base while healing

Gireen:
fix Unvanquished/Unvanquished#1211


slipher:
Looks good. FWIW eggs also have a healing effect equal to the overmind's.

Gireen:
I think any alien building does, or more precise its creep . But bots only use the booster and overmind to heal.

PR #31, opened 2020-12-29
Fix stuck bot

necessarily-equal:
Screenshot from 2020-12-29 03-21-56

That bot stayed stuck for like, 5 minutes because he was still close to the turret he wanted to fix originally.

### DIFF
--- a/bots/default.bt
+++ b/bots/default.bt
@@ -4,7 +4,7 @@ selector
 	{
 		sequence
 		{
-			condition haveWeapon( WP_HBUILD ) && distanceTo( E_GOAL ) <= 100
+			condition haveWeapon( WP_HBUILD ) && distanceTo( E_GOAL ) <= 100 && percentHealth( E_GOAL ) < 1.0
 			action resetStuckTime
 		}
 	}


slipher:
`buildingIsDamaged` is true when there is any damaged building on the entire map. So I doubt this change would have prevented the problem you observed.

The change looks correct though, in the cases where it applies. It would be good to change the message to reflect the fact that it only applies when there are no damaged buildings.

necessarily-equal:
I agree that it's not really good yet, even if it's still an 
improvement. Do you know if it is possible to query if a specific 
building is damaged?

slipher:
I don't see anything implemented for checking the health of an arbitrary entity in the `AIActions` table of sg_bot_parse.cpp

Looking at the `BotActionRepair` code in sg_bot_ai.cpp, if the building healed on its own, that should be treated no differently than if the bot managed to repair it. If at any time the target reaches full health (or dies), the action finished.

One thing that might possibly be happening: every alternative in the base selector fails, and the bot does not attempt any new action. This could happen if there is some issue with pathfinding, since all the actions depend on being able to find a path to some target. If this happens, I guess the goal entity would still be stuck on whatever the last goal was, which could result in the stuck-detection being disabled in the way you hypothesize. Maybe at the end of the main selector (reached only if all alternatives failed), we could set the goal to null or something. Not sure if there's already a way to do that.

Gireen:
`percentHealth( E_GOAL )` could be used to get the buildings health 

But i don't completely get what happened here. Is the stuck bot the one left or do we see his perspective?
Was the bot standing there or trying to move? Was he trying to heal the turret but couldn't reach it or done and just standing there? 

It would be an oddly specific situation if @slipher is right. A bot would have to leave the navmesh close to a building it wants to repair but the building itself has to be on said navmesh that the bot can reach there in the first place. 😄

> Maybe at the end of the main selector (reached only if all alternatives failed), we could set the goal to null or something. I think this could be done with the changeGoal action.

that is in every case a good change.

necessarily-equal:
So, to explain a bit more, the stuck bot is the one on the left. It managed to get up there because when alien attacked the base, it switched to the blaster and started fighting. For some reason it jumped and ended up up there. My guess is that the condition that blocked the "unstucking procedure" triggered because the bot was still very close to the turret it originally wanted to repair.

slipher:
Oh, I assumed the bot was the one in first-person view since it also has a ckit. The one on the side really does look likely to be off the navmesh.

> For some reason it jumped

This may happen due to the "unstick" code...

necessarily-equal:
I've pushed a new version that only considers the target building and with a shorter distance. Hopefully, it won't prevent the unstuck code from running.

necessarily-equal:
I've reverted the distance change, because it cause an issue when the bot is higher than the building it's reparing.

illwieckz:
@necessarily-equal what is the status on this?

illwieckz:
Hmm, the change seems to be fairly simple and safe, somewhat given wounded buildables are either healing themselves or are dead…

PR #33, opened 2020-12-30
 Allow visualisation of 2× healing

necessarily-equal:
For some reason, git forgot that this is actually:

rename 2.png -> 3.png
rename 3.png -> 4.png
add a new 2.png

to be merged along https://github.com/Unvanquished/Unvanquished/pull/1251.


illwieckz:
I merged it by hand after having optimized the png (going from `4346 B` to `145 B`).
Thank you very-much!

PR #35, opened 2021-01-06
Cleaner alien evo cost

necessarily-equal:
before:
	Price: 2.0 morph points
	Price: -1.8 morph points

after:
	Price: 2.0 morph points
	Returned: 1.8 morph points


illwieckz:
Interesting ! 👀

necessarily-equal:
This should be merged now that https://github.com/Unvanquished/Unvanquished/commit/f489d156debc3e86d3b4a509e34d86bf3b8a026c is in Unv

illwieckz:
Right

PR #36, opened 2021-02-09
some updates on bult-in gameplay help

illwieckz:
some updates on bult-in gameplay help

### DIFF
--- a/ui/help_gameplay.rml
+++ b/ui/help_gameplay.rml
@@ -23,40 +23,95 @@
 				Welcome
 			</tab>
 			<panel>
-				In this section you will find a summary of major gameplay changes since Tremulous GPP.
+				In this section you will find a summary of major gameplay
+				elements to know to become a successful player.
 				<br />
 				<br />
 				Feel free to take part in our weekly development games, each
 				<span clas="red">Sunday</span>
 				at
-				<span class="red">20:00 UTC</span>
-				on the »Dev &amp; Testing Server«!
+				<span class="red">20:00 UTC</span>!
+			</panel>
+			<tab>
+				<img class="emoticon" src="/emoticons/tyrant" />
+				Income &amp; evolution
+			</tab>
+			<panel>
+				<img class="emoticon" src="/emoticons/dretch" />
+				Aliens can earn evolution points by killing humans
+				and destroying human buildables,
+				then they can use those evolution points to evolve into
+				<img class="emoticon" src="/emoticons/dragoon" />
+				more advanced forms.
+				<br />
+				<br />
+				Aliens can evolve from everywhere if not in line of sight
+				of an human player or buildable. Aliens can also devolve to
+				<img class="emoticon" src="/emoticons/granger" />
+				less advanced forms (at some cost) when being near the
+				<img class="emoticon" src="/emoticons/overmind" />
+				overmind.
+				<br />
+				<br />
+				Build an overmind with a
+				<img class="emoticon" src="/emoticons/granger" />
+				granger if missing, you need a live overmind to evolve!
+				<br />
+				<br />
+				<img class="emoticon" src="/emoticons/human" />
+				Humans can earn credits by killing aliens and destroying alien buildables,
+				then they can uses those credits to buy new
+				<img class="emoticon" src="/emoticons/lcannon" />
+				weapons, body armours or
+				<img class="emoticon" src="/emoticons/bsuit" />
+				battlesuit, and accessories like a radar or a jetpack.
+				<br />
+				<br />
+				Humans buy and sell weapons at the
+				<img class="emoticon" src="/emoticons/armoury" />
+				armoury, make sure one is built! Use the construction kit
+				to build an armoury if it's missing.
 			</panel>
 			<tab>
 				<img class="emoticon" src="/emoticons/drill" />
-				Resources
+				Powering buildables
 			</tab>
 			<panel>
-				Building a structure requires build points, which are now a one-time
-				payment – if the buildable gets destroyed, the build points spent on it
-				are lost.
+				Like the
+				<img class="emoticon" src="/emoticons/overmind" />
+				overmind the
+				<img class="emoticon" src="/emoticons/reactor" />
+				reactor now supplies power to every place on the map, but
+				there is a limited amount of energy per region. Human buildables will
+				compete for power and shut down if they don't get enough.
 				<br />
 				<br />
-				Your team will start with a small pool of build points but more can be
-				acquired by building a mining structure.
+				When the power consumption in a region is too high, the smallest possible
+				number of buildables will power down so that the other structures have
+				enough energy to remain operational.
+				<br />
+				<br />
+				Building a structure requires build points. Your team will start with a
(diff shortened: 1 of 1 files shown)


ui/help_gameplay.rml
```diff
@@ -37,26 +37,44 @@
...
 				<img class="emoticon" src="/emoticons/drill" />
 				drill and
 				the alien
 				<img class="emoticon" src="/emoticons/leech" />
-				leech. These buildings need to be spread out for
-				efficiency.
+				leech. These buildings need to be spread out for efficiency.
+				<br />
+				<br />
+				Don't build mining structures too close together or they will compete with each others
```
slipher, line 76: s/others/other

ui/help_gameplay.rml
```diff
@@ -122,7 +122,7 @@
 				The
 				<img class="emoticon" src="/emoticons/advgranger" />
 				advanced granger's spit attack can be used to put out
-				fires.
+				fires or slow downs intruders!
```
slipher, line 125: s/downs/down/

ui/help_gameplay.rml
illwieckz: done

ui/help_gameplay.rml
illwieckz: done

illwieckz:
I added more text to that gameplay guide.

ui/help_gameplay.rml
```diff
@@ -23,40 +23,98 @@
...
+				there is a limited amount of energy per region. Human buildables will
+				compete for power and shut down if they don't get enough.
+				<br />
+				<br />
+				When the power consumption in a region is too high, the smallest possible
+				number of buildables will power down so that the other structures have
+				enough energy to remain operational.
+				<br />
+				<br />
 				Building a structure requires build points, which are now a one-time
```
slipher, line 94: I think this is incorrect now... the BP pool is just determined by the mining structures, right?

ui/help_gameplay.rml
```diff
@@ -131,45 +155,115 @@
...
+				<br />
+				<img class="emoticon" src="/emoticons/dretch" />
+				Aliens heal slowly whatever they are, but they heal faster
+				in groups and near alien buildables. Building forward ones
+				can be decisive!
+				<br/>
+				<br/>
+				The
+				<img class="emoticon" src="/emoticons/booster" />
+				Booster is the fastest healing buildable, build one as soon
```
slipher, line 202: The wording here is not the clearest, how about "The booster gives rapid healing for nearby aliens"

ui/help_gameplay.rml
```diff
@@ -131,45 +155,115 @@
...
-				The human dodge has been removed, stamina has been increased.
+				Since Tremulous GPP The human dodge has been removed,
+				stamina has been increased.
+				<br />
+				<br />
+				All alien evolution costs and gains have been doubled.
+				The new maximum of evos an alien can store is 20 (formerly 9).
+				<br />
+				<br />
+				Humans no longer have the Tesla defense tower neither the defense
```
slipher, line 257: s/neither/nor/

ui/help_gameplay.rml
illwieckz: No, the mining structures continue to mines resources (with the rate decreasing over time). And in all case, the BP are lost.

Let's imagine I build a drill, let's say it enable 40 bp (I don't remember the exact number, but that's close) and mines 20 other bp. So I have 60 bp.

After I build 6 turrets which costs 10 bp each. So I spent all the bp.

4 turrets have been destroyed but the drill is still alive, I need to find 40bp again to rebuild those 2 turrets, for example by building a second drill.

Then the next 2 turrets have been destroyed as well, I need to find 20bp to rebuild those 2 turrets.

If I loose a drill, I will lose 40bp and I will miss 40bp, some buildable will disable themselves, rebuilding the drill would just make me miss no bp, I would gain no bp by building it.

That's why at some point, some team ends with -100 bp and they can't build anything even by spamming drills because they lost too much bp and they would never be able to build enough drills to recover, this replace sudden death.

ui/help_gameplay.rml
illwieckz: I applied the suggestion.

ui/help_gameplay.rml
illwieckz: done

ui/help_gameplay.rml
slipher: The build points are not permanently lost, but rather put into the "recovery queue". The BP exit the queue one at a time. The queue processes BP at an initial rate of 16/minute [g_BPRecoveryInitialRate]; this rate halves every 10 [g_BPRecoveryRateHalfLife] minutes.

illwieckz: I'm not sure to understand everything, so I made this part of the help _more vague_, anyway there is enough information for the player to know how he can lose and earn build points.

ui/help_gameplay.rml
```diff
@@ -131,45 +152,115 @@
...
+				<br />
+				The advanced granger can be unlocked with momentum.
+			</panel>
 			<tab>
 				<img class="emoticon" src="/emoticons/dev" />
 				Miscellaneous
 			</tab>
 			<panel>
-				The human dodge has been removed, stamina has been increased.
+				Since Tremulous GPP The human dodge has been removed,
```
slipher, line 246: Let's drop this paragraph since GPP is hardly what anyone will remember best nowadays.

ui/help_gameplay.rml
```diff
@@ -23,40 +23,95 @@
...
+				When the power consumption in a region is too high, the smallest possible
+				number of buildables will power down so that the other structures have
+				enough energy to remain operational.
+				<br />
+				<br />
+				Building a structure requires build points. Your team will start with a
+				small pool of build points but if you want to place a greater number of
+				buildables in an area, you can build a
+				<img class="emoticon" src="/emoticons/leech" />
+				mining structure to increase the available power and extract more build points.
```
slipher, line 98: To me "extract" sounds like the build points would be stored permanently. I suggest "provide"

ui/help_gameplay.rml
illwieckz: Was human dodge GPP specific (not in 1.1)?

slipher: Yes

illwieckz:
I updated the PR. This PR does not aim to produce a perfect gameplay guide or this will never be ready, but at least fix some of the biggest issues of the former one that was not updated since probably five years or more. Like, if the problem is that Tremulous GPP is mentioned in the last chapter of the guide, it means the guide is better because previously that mention was the introduction and it was not only required for the player to remind GPP, but also remind all the details of GPP… So, I fixed that GPP mention but I hope you got it. Most remarks that can be done about the current state of the guide is not anymore about fixing the former guide, but about taking this guide as a reference to improve it as a next step. So it's for another PR anyone can submit once this one is merged.

illwieckz:
The various remarks has been addressed. Other changes are welcome in other PR.

PR #37, opened 2021-02-11
Don't show devolving as possible when far from OM

necessarily-equal:
comes with https://github.com/Unvanquished/Unvanquished/pull/1320.


illwieckz:
Is there a message telling why it's disabled? Because if we can't click on it we will not get the message saying why.

necessarily-equal:
That's a fair point, and that's a problem with every message. Maybe we 
should set `action` for every line anyhow

PR #38, opened 2021-02-22
Move emoticon SVG source into different directory

slipher:
See https://github.com/Unvanquished/Unvanquished/pull/1335


illwieckz:
Hmm, I'm not really happy with this, in other repositories xqf sources are stored on the same folder, and we can probably find some other svg sources being stored next to the exported bitmap ones. I guess that's required because the list of emoticons is dynamically built from any basename in such folder?

slipher:
What's xqf - that's not an image format is it?

This change affects only emoticons, not anything else. Emoticons have this kind of requirement because we would like to be able to tell which emoticons are available in sgame also, where there is no image-loading machinery available.

I checked `res-legacy_src.dpkdir` and there are only PNGs there.

illwieckz:
Sorry, I wanted to say `xcf` (GIMP source).

You're right about sgame, so let's do this exception!

PR #39, opened 2021-03-11
split default behavior tree into species specific files

bmorel:
the splits makes it easier to understand either aliens or humans behavior, in the hope to improve them.
It also revealed some refactoring which result in less nodes, by merging e.g.

```
sequence
{
  condition test1
  condition test2
  action do_something
}
```

into:

```
condition test1 && test2
{
  action do_something
}
```

Which should be more obvious to most people.

### DIFF
--- a/bots/default.bt
+++ b/bots/default.bt
@@ -1,107 +1,11 @@
 selector
 {
-	decorator return( STATUS_FAILURE )
-	{
-		sequence
-		{
-			condition haveWeapon( WP_HBUILD ) && distanceTo( E_GOAL ) <= 100
-			action resetStuckTime
-		}
-	}
-
-	behavior unstick
-
-	sequence
-	{
-		condition alertedToEnemy
-		selector
-		{
-			condition haveWeapon( WP_HBUILD ) && ( !buildingIsDamaged || teamateHasWeapon( WP_HBUILD ) )
-			{
-				selector
-				{
-					action equip
-					action flee
-				}
-			}
-
-			decorator timer( 1000 )
-			{
-				selector
-				{
-					sequence
-					{
-						condition alertedToEnemy
-						condition team == TEAM_ALIENS
-						condition ( distanceTo( E_A_OVERMIND ) <= 100 || distanceTo( E_A_BOOSTER ) <= 100 )
-						action fight
-					}
-					sequence
-					{
-						condition ( healScore <= 0.5 && percentHealth( E_GOAL ) > 0.3  && random > 0.3 )
-						action heal
-					}
-				}
-			}
-
-			action fight
-		}
-	}
-
-	selector
-	{
-		sequence
-		{
-			condition team == TEAM_ALIENS
-			condition healScore <= 0.25
-			action heal
-		}
-
-		sequence
-		{
-			condition team == TEAM_HUMANS
-			condition !haveUpgrade( UP_MEDKIT )
-			condition healScore <= 0.25
-			action heal
-		}
-	}
-
-	condition team == TEAM_ALIENS
-	{
-		action evolve
-	}
-
-	sequence
-	{
-		condition team == TEAM_HUMANS
-		condition !teamateHasWeapon( WP_HBUILD )
-		condition buildingIsDamaged
-		decorator timer( 50000 )
-		{
-			selector
-			{
-				condition !haveWeapon( WP_HBUILD )
-				{
-					action buy( WP_HBUILD )
-				}
-
-				condition haveWeapon( WP_HBUILD )
-				{
-					action repair
-				}
-			}
-		}
-	}
-
 	condition team == TEAM_HUMANS
 	{
-		action equip
+		behavior defaultHumans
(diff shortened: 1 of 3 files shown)


illwieckz:
Oh yes that's clearly more readable that way! And allow to properly focus on a team for implementing improvements (less mental load). :smiley:

bots/default.bt
```diff
@@ -1,107 +1,11 @@
...
 	{
-		action equip
+		behavior humans
 	}
-
-	condition baseRushScore > 0.5
+	condition team == TEAM_ALIENS
 	{
-		action rush
+		behavior aliens
```
necessarily-equal, line 9: maybe better name the behaviours `defaultAlien` and `defaultHuman`.

necessarily-equal (changes requested):
I like how it's not even longer

bots/default.bt
slipher: +1

bots/humans.bt
```diff
@@ -0,0 +1,65 @@
...
+	condition !haveUpgrade( UP_MEDKIT ) && healScore <= 0.25
+	{
+		action heal
+	}
+
+	condition !teamateHasWeapon( WP_HBUILD ) && buildingIsDamaged
+	{
+		decorator timer( 50000 )
+		{
+			sequence
```
slipher, line 47: This part changed compared to the old version

slipher (commented):
Overall this is a good simplification.

bots/humans.bt
bmorel: please tell me more? Or better, show me how I accidentally changed behavior?

slipher: changed selector to sequence and deleted `condition haveWeapon( WP_HBUILD )`

bmorel: Exact.
Previous behavior was: "if I do not have ckit, buy one, ELSE repair" where current code says "if I do not have ckit, buy one. repair anyway."
Do you see any reason to keep old algorithm? Or maybe I misunderstand it?

slipher: It is not the same since the bot may decide to do something else instead of repairing, after buying the ckit.

I don't know if the changed version is better or worse, but I thought the point of this commit is to do a non-behavior-changing refactor.

bmorel: > but I thought the point of this commit is to do a non-behavior-changing refactor.

True it was, but I feel like it was a bug: why would a bot take a ckit if not to repair? They do not know how to build (yet) and ckit can only be used to do one or the other.

necessarily-equal:
If you want to introduce unrelated changes, they belong in another 
commit IMO. Otherwise that's fine

bots/humans.bt
slipher: One possibility is that if after the bot buys a ckit, if it sees an enemy and someone else already has a ckit, it will change plans to buy a gun and try to attack the enemy.

bmorel: I'm not really convinced, but it's not really important anyway, I'll restore previous behavior.

slipher:
End result look OK, just waiting for a rebase (so files aren't renamed multiple times etc.)

Personally I would much rather see the rebases done right away instead of delayed as late as possible. I usually look at the individual commits (isn't that what they're there for?) rather than the overall diff, so it's annoying if commits with mistakes that have already been pointed out stay in.

bmorel:
> I usually look at the individual commits (isn't that what they're there for?)

Indeed, I also consider commits a may better source of information than a simple diff (and I don't like the empty commits introduced by github when merging, neither, because they just don't make sense since other commits of the merge can be interleaved with commits of other merges). But I find it easier to address first all issues and *then* rebase to have meaningful commits (less rebases less bugs introduced while moving things before and after and fixing the resulting conflicts).
The 3 commits here (should be only 3, one should not have been here... sigh) can easily be squashed thought.

slipher:
The [problem](https://github.com/Unvanquished/Unvanquished/issues/1374) may have been that
```
sequence
{
  condition test1
  action do_something
}
```
and
```
condition test1 
{
  action do_something
}
```

are NOT equivalent when `do_something` is a long-running action. The former evaluates the condition only once, while the latter evaluates it every frame.

bmorel:
I don't remember, honestly. I only know that I have a working solution since months. You can see a part of it here: https://github.com/bmorel/unvanquished_src.dpkdir/commit/7dd908000c5880b09ec8c674c40af931e9dd051e (more recent commits are not available, github is not really sexy, always requiring me to get mail and write back a stupid code...)

illwieckz:
> (more recent commits are not available, github is not really sexy, always requiring me to get mail and write back a stupid code...)

Hmm, do you know that you can [add a SSH public key](https://github.com/settings/keys) to GitHub, then use `git@github.com:bmorel/unvanquished_src.dpkdir.git` remote url? Or do I am missing something?

bmorel:
I know that, I just mostly sync stuff when I need to do a PR

PR #40, opened 2021-03-17
Circle menu rewrite (number key support)

slipher:
Depends on https://github.com/Unvanquished/Unvanquished/pull/1348.

UX improvements to the circle menu:
- Position in the center of the screen, so that the cursor appears in the center (as suggested by @necessarily-equal)
- Add the ability to select items with number keys, for faster usage

unvanquished-beaconmenu-nums

For the moment I have applied the changes only to the beacons circle menu, which can be opened by `/beaconmenu`.

Additionally, the circle menu code is generated in Lua instead of C++. Besides being simpler/cleaner in my opinion, this way is much better for development in a couple ways:
- Changes can be tested by just reloading the UI (e.g. `/vid_restart`), without recompiling
- Changes can be mostly made in this repository, rather than a mixture of both it and `Unvanquished`.

### DIFF
--- a/ui/alien_build.rml
+++ b/ui/alien_build.rml
@@ -6,7 +6,6 @@
 <style>
 body
 {
-	padding-top:10%;
 	width:100%;
 	height:100%;
 	background:#00000066;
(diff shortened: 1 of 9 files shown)


illwieckz:
Wow that's nice!!!! Would this make easier to implement gamepad support one day, or that's unrelated?

slipher:
> Would this make easier to implement gamepad support one day, or that's unrelated?

Future modifications to the code should be easier, if nothing else.

To use the code directly we would need to have 1 controller button mapped for each of up to 10 options, but this seems clunky; it seems nicer to move a stick towards the angle of the desired option or something.

illwieckz:
> Future modifications to the code should be easier, if nothing else.

That's good! Design making other things easier is good design!

> To use the code directly we would need to have 1 controller button mapped for each of up to 10 options, but this seems clunky; it seems nicer to move a stick towards the angle of the desired option or something.

For a start, right/left to iterate the options with one key to select the entry and one key to quit the menu would be a working alternative (even it moving toward the angle would be the best one, but probably far more complicated to implement).

slipher:
This is now ready. There's one thing I'm thinking about adding, which is to switch between weapons and upgrades (at the armoury) by pressing <kbd>TAB</kbd>.

illwieckz:
LGTM.

I noticed than, when you're advanced granger but lost momentum so advanced granger is locked, you can devolve from advanced granger to granger, but that's probably unrelated to this.

I've seen nothing wrong anyway.

Tab key to switch between circle menu tabs would be great!

necessarily-equal:
Hi, two points I noticed:

Screenshot from 2021-04-03 10-48-32  

1. The build menu has been rotated by two elements, so that the drill that was on top is now on the left and the telenode that was on the right is now on top
2. At least on low res screens, the border around the numbers is not pretty

Otherwise, looks good. having a key like `tab` to switch to and fro armors would be nice indeed

slipher:
> 1. The build menu has been rotated by two elements, so that the drill that was on top is now on the left and the telenode that was on the right is now on top

Is there some reason to prefer one order over the other?

> 2. At least on low res screens, the border around the numbers is not pretty

I know :cry: Hopefully someone with more artistic skills can fix it someday.

slipher:
Implemented the TAB idea.

unvanquished-humanbuy-tab

necessarily-equal:
> Is there some reason to prefer one order over the other?

muscle memory :P

illwieckz:
This is really really good! 😯 I don't mind the default position has changed, we can relearn, and the community is small anyway. 😏

Now that makes me think that if it was possible to iterate the buttons with the mouse wheel (with the mouse pointer following the highlighted circle), one would not have to move the mouse at all: open menu, mouse wheel steps, click.

PR #42, opened 2021-04-08
presets: enable fastsky with lowest graphics

illwieckz:
Also enable GL Clearing with all presets.

Also reword a bit the description of the `lowest` preset, not mentioning there is no multitexturing (`low` does not have it as well).


slipher:
LGTM

PR #45, opened 2021-04-12
Reimport .attr.cfg, .locdamage.cfg and .missile.cfg files, replay modifying commits

illwieckz:
Note: `.missile.cfg` files has path to files in `res-weapons`, it would be very good to split those the same way weapons have `.attr.cfg` and `.model.cfg` files. So half of their content has to live in `res-weapons` to avoid having to edit two repositories when renaming a texture file or replacing a model…

See https://github.com/Unvanquished/Unvanquished/issues/1352


illwieckz:
ping @bmorel @necessarily-equal @Gireen 👀

illwieckz:
Well, for missiles it looks like we can exactly split `.missile.cfg` files into `attr.cfg` and `.model.cfg` ones, because the content is just about attributes and models (like weapons, buildables, players)…

illwieckz:
See:

- https://github.com/UnvanquishedAssets/res-players_src.dpkdir/pull/10
- https://github.com/UnvanquishedAssets/res-buildables_src.dpkdir/pull/6
- https://github.com/UnvanquishedAssets/res-weapons_src.dpkdir/pull/8

illwieckz:
See: https://github.com/Unvanquished/Unvanquished/pull/1367

Gireen:
Could we move this directly into the unvanquished repository? 
To quote myself

> maybe renaming unvanquished_src.dpkdir to res-basics_src.dpkdir containing ui image assets and the like
and create a new unvanquished_src.dpkdir in Unvanquished/Unvanquished containing all config, rml and bt files

illwieckz:
It's too early to merge them in Unvanquished repository (if we can do it), because it brings a load of new problems to solve related to the way .dpk archives are built very differently than binaries.

PR #46, opened 2021-04-19
SHIFT ENTER for teamchat

Gireen:


illwieckz:
That's a good idea.

That would require a similar fix being done on other presets (note: I just pushed a commit because I noticed I forgot the ENTER chat bind in one, so you need to rebase).

It would require to edit the Key UI accordingly, similarly to that:

https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/blob/6585c629709202f049f2a623dc557faf300c17c9/ui/options_keys.rml#L163-L167

illwieckz:
Hmmm, did you properly rebase? it saids there is a conflict.

PR #49, opened 2021-04-24
bots: add aliveTime condition

illwieckz:
Make sure bots wait a little (currently `1.5s`) before trying to evolve after spawn, given eggs are usually hidden in small places, the surface where the egg is may not be disconnected as dretch, but may be disconnected as dragoon.

See https://github.com/Unvanquished/Unvanquished/pull/1310

### DIFF
--- a/bots/camper.bt
+++ b/bots/camper.bt
@@ -30,7 +30,10 @@ selector
 			{
 				action heal
 			}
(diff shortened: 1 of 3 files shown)


illwieckz:
In game testing shown it works.

PR #53, opened 2021-05-03
remove no-effect cVar UI to match unvanquished similar patch

bmorel:


illwieckz:
I modified target to `master` now that we have merged `0.52.0/sync` to `master`, maybe a rebase is needed.

illwieckz:
@slipher maybe we can merge this right now even if we postpone merging https://github.com/Unvanquished/Unvanquished/pull/1400 for after 0.52.0 (because of code freeze.). I highly doubt such UI change can introduce a regression, right?

slipher:
> 
> 
> @slipher maybe we can merge this right now even if we postpone merging [Unvanquished/Unvanquished#1400](https://github.com/Unvanquished/Unvanquished/pull/1400) for after 0.52.0 (because of code freeze.). I highly doubt such UI change can introduce a regression, right?

The risk is not huge, but the benefit is also small. I don't favor throwing in more last-minute stuff if it is not so important.

bmorel:
I agree that there is no reason to accept it before a release. I'd prefer to see them integrated later, and there are other cvars to remove that I noticed today (they don't have UI though) while other bot related have effect but not UI. Maybe better a single PR with removal of all bot cvars I know are not used, and which push UI for the few ones I've seen which have effect?

bmorel:
oops, this was using master branch that I synced with less changes. I'm closing it since that's in my betterai branch anyway, which contains various other changes regarding that area too.

PR #55, opened 2021-05-28
use hw:c by default to taunt in wasd presets

bmorel:
0.51 used 'c' key, which have only advantages over the new hw:[ key:

* closer to directional keys (wasd/zqsd depending on layout)
* same key in all qwerty and azerty layouts I'm aware of (not many, but
  still)
* less surprising to people which have played to 0.51 (since old
  defaults)

closes #1405


illwieckz:
Note: it looks like the original problematic change was a merge issue and was not intentional.

PR #56, opened 2021-06-01
presets/graphics: re-set r_clear to 0

illwieckz:
See:

- https://github.com/DaemonEngine/Daemon/issues/472
- https://github.com/DaemonEngine/Daemon/pull/473

### DIFF
--- a/presets/graphics/high.cfg
+++ b/presets/graphics/high.cfg
@@ -14,7 +14,7 @@ r_imageMaxDimension 2048
 r_ext_multisample 8
 r_ext_texture_filter_anisotropic 8
 r_textureMode "GL_LINEAR_MIPMAP_LINEAR"
-r_clear 1
+r_clear 0
 r_fastsky 0
 cg_shadows 1
 cg_playerShadows 1
--- a/presets/graphics/low.cfg
+++ b/presets/graphics/low.cfg
@@ -14,7 +14,7 @@ r_imageMaxDimension 256
 r_ext_multisample 2
 r_ext_texture_filter_anisotropic 2
 r_textureMode "GL_LINEAR_MIPMAP_NEAREST"
-r_clear 1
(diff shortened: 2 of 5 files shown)


slipher:
Well we have a fix for for the r_clear-triggered bug, but maybe we should still set it to 0 for the non-fastsky ones as a precaution against other bugs?

illwieckz:
After some thoughts, this is what I think about:

- set `r_clear 1` for usual profiles (medium, high, ultra), on today's hardware player should not see any difference, really, and this would avoids wall of mirrors and other issues from third-party maps we can't fix.
- set `r_clear 0` on low and lowest profile, to save a potential extra fps, making use of https://github.com/DaemonEngine/Daemon/pull/473 to clear only when there is sky on the scene and fastky is enabled. I have seen a gain doing that an a PCI (cough, cough) Tesla (cough) card from 2007 on a Vista-era PC (cough, cough)… So it makes sense that if someones _needs_ lowest preset, such trick would be appreciated.

I'm yet to see some hardware supporting medium preset to get a performance difference from `r_clear 0`, so better fix bugs from legacy third-party assets starting with medium presets.

slipher:
Daemon#473 seems doubtful to me. `Tess_StageIteratorSky` is part of the surface drawing pipeline. Clearing of course needs to be done before anything is drawn. So clearing the screen after surfaces are already being drawn risks erasing things.

illwieckz:
But isn't sky meant to be drawn before anything else? As far as I know that's part of why we can't draw sky over other things and then, sometime objects from other rooms can be seen through the sky… because sky is meant to be drawn before.

slipher:
In the normal case I suppose the sky should always have the largest depth and be sorted first. But I'm not sure it couldn't happen in some case like with cameras/portals etc.

PR #57, opened 2021-06-11
Add different options to close evolve circle menu

Gireen:
cg_closeCircleMenu = 0 default behavior
1 = any key can close the menu
2 = key must be hold to keep menu open

### DIFF
--- a/ui/alien_evo.rml
+++ b/ui/alien_evo.rml
@@ -117,6 +117,7 @@ tabs {
 	end
 	function BuildEvolveMenu(document)
 		ClearEvolveInfo(document)
+		CirclemenuSetKeys(document)
 		CDataSource.Build("alienEvolveList", "default")
 		evolutions = CDataSource.Read("alienEvolveList", "default", "icon,name,description,availability,price,visible")
 		button_to_index = {}
@@ -137,7 +138,8 @@ tabs {
 	end
 </script>
 </head>
-<body id="alien_evo" class="circlemenu alien" onkeydown="CirclemenuHandleKey(event, document, EvolutionClick)" onshow="BuildEvolveMenu(document)">
+<body id="alien_evo" class="circlemenu alien" onkeydown="CirclemenuHandleKey(event, document, EvolutionClick)"
+ onkeyup="CirclemenuHandleKeyUp(event, document)" onshow="BuildEvolveMenu(document)">
 <tabset class="circlemenu">
 <tab></tab>
 <panel>
@@ -149,5 +151,6 @@ tabs {
 </div>
 </panel>
 </tabset>
+<keybind id="circlemenu-binds" style="display:none;overflow:hidden;widht:0px;height:0px;" cmd="+activate" team="aliens" />
 </body>
 </rml>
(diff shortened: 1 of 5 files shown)


slipher:
Mode 1 seems like a good idea. Why not have mode 1 trigger on key down? Then there is no need for the global variable recording whether a key was pressed. This way putting the code in `CirclemenuHandleKey`, it is also easy to avoid closing if the key pressed was a number or tab (important for the armory menu).

Mode 2 is kinda janky to my taste, since the menu goes away with any keyup, not only the one opening the menu (for example if you press the evolve menu while running). Also, it might be hard to make this mode work with the armory menu. But we could add it if people want it.

Gireen:
Indeed its more complicated than necessary. 
Ideally everything would only react on keys bound to +activate. 
It would be possible to hardcore it to `Q` but that's not so nice.

illwieckz:
Two thoughts:

If one day we want to implement gamepad support in UI, we may want to be able to cycle the items with keys (like previous/next actions), so actual behaviour would be good.

With “key must be hold to keep the menu open”, we may still send the “wallwalk key” while the menu is open, so player can enable non-toggled wallwalk and not fall when opening the build menu.

slipher:
> If one day we want to implement gamepad support in UI, we may want to be able to cycle the items with keys (like previous/next actions), so actual behaviour would be good.

What is "actual behaviour"?

> With “key must be hold to keep the menu open”, we may still send the “wallwalk key” while the menu is open, so player can enable non-toggled wallwalk and not fall when opening the build menu.

Currently if you are holding down a key bound to a +command when you close the menu, the command won't resume sending. I guess that could be considered a bug, which would need to be fixed first.

illwieckz:
> What is "actual behaviour"?

One key to open the menu, the menu being kept open until another key closes it.

Gireen:
After a few test i found that this version is already the best possible at the moment.
For me its good enough. There are no other buttons to be pushed in the alien menu and all other menus are more complex anyway.

For mode 1 its necessary to listen on keyup cause the keydown event is trigger often if a key is hold down.
There is no big interference with function of the 0-9 keys except that the menu closes if a action is not possible.

> If one day we want to implement gamepad support in UI, we may want to be able to cycle the items with keys (like previous/next actions), so actual behaviour would be good.

Gamepads can use the joystick and if not then we can add an exception here anytime

>With “key must be hold to keep the menu open”, we may still send the “wallwalk key” while the menu is open, so player can enable non-toggled wallwalk and not fall when opening the build menu.

Its not possible to evolve while wallwalking. Otherwise that problem should have its own issue.

illwieckz:
> Gamepads can use the joystick and if not then we can add an exception here anytime

Currently, a gamepad player has no way to select any option neither close a GUI once he opened one.

> Its not possible to evolve while wallwalking. Otherwise that problem should have its own issue.

But it is definitely possible to build while wallwalking.

slipher:
> For mode 1 its necessary to listen on keyup cause the keydown event is trigger often if a key is hold down.

True, there are key repeat events designed for text entry (since librocket is using the character key events rather than the scancode ones). But this issue also affects mode 1. If you hold down the key used to open the menu (or any other key), then it will start generating key down events, and when you let go it will close the menu, despite you not having done any real keydowns since opening the menu.

Gireen:
Now it should work perfectly with everything except build menus

slipher:
We finally implemented this :) https://github.com/Unvanquished/Unvanquished/pull/2900

PR #58, opened 2021-07-24
gives (most) alien blobs some volume

bmorel:
After few discussions on IRC today, I got a look at alien's missiles sizes. Most have a size of 0, which the code interprets as something very small. I think this is fine for the Hive, since bees should be small, but for blobs, I think it just makes it harder to use them and that might not be intended.
This commit sets those sizes at 5, which is the usual size: another numbers might be better choices.

### DIFF
--- a/configs/missiles/bounceball.attr.cfg
+++ b/configs/missiles/bounceball.attr.cfg
@@ -7,7 +7,7 @@ splashRadius       75
 splashMeansOfDeath MOD_LEVEL3_BOUNCEBALL
 
 clipmask           MASK_SHOT
-size               0
+size               5
 
 trajectory         TR_GRAVITY
 speed              1000
--- a/configs/missiles/lockblob.attr.cfg
+++ b/configs/missiles/lockblob.attr.cfg
@@ -3,7 +3,7 @@ damage             0
 meansOfDeath       MOD_UNKNOWN
 
 clipmask           MASK_SHOT
-size               0
+size               5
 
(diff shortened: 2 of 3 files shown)


slipher:
I guess it should correspond to the size of the model? The blaster is 5 but its bbox seems too big.

necessarily-equal:
We basically need to pass over each missile size and tweak :S

bmorel:
I can go through all models to check the sizes, if that's the only "con" preventing this PR to be merged.
Now, remember that blaster is hard enough to use as it is, so I suspect reducing the bolt's BBox might be quite the nerf. I'm not fond of the idea of nerfing blaster...

necessarily-equal:
I'd say that the cons, IMO, is that the goon blob ("bounceball") still feels a bit small. maybe the granger one too?

Le 6 novembre 2021 14:54:33 GMT+01:00, "Morel Bérenger" ***@***.***> a écrit :

bmorel:
Those are, like, improved cons, no? That is, it's still bad, but also still better :p

To check about that in-game, though, what you need to do is to reduce time scale drastically, enable bbox drawing and... check. I don't remember having tested it tbh, but I'm fine with increasing the BBoxes of spits.

bmorel:
I won't merge.

necessarily-equal:
I mean, it's indeed better. The thing with the blob size is more like "it would be better if" and non-blocking IMO

PR #60, opened 2021-09-27
add build timer

Gireen:
required for buildqueue

### DIFF
--- a/ui/hud_builder.rml
+++ b/ui/hud_builder.rml
@@ -23,9 +23,10 @@ div#timer
 	position: absolute;
 	width: 40%;
 	height: 0.8em;
-	bottom: 0%;
+	bottom: 3%;
 	margin-left: auto;
 	margin-right: auto;
+	text-align:center;
 }
 div#rate
 {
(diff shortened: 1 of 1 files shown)


bmorel:
This is a PR to go with the build queue in unvanquished repo, I assume?

Gireen:
Correct. it replaces the progress bar with a text and timer how much additional build time is added.

ui/hud_builder.rml
```diff
@@ -43,22 +44,29 @@ progressbar
...
+text-align:center;
+}
 
 .warning {
 	color: yellow;
 }
 
 .error {
 	color: red;
-}
```
bmorel, line 53: where is the '}' gone?

ui/hud_builder.rml
```diff
@@ -43,22 +44,29 @@ progressbar
 	image: /gfx/colors/white;
 	orientation: left;
 }
+#buildQueueTime
```
bmorel, line 47: looks like there's an empty line for nothing and content is not indented

PR #61, opened 2021-10-01
Remove client translation messages

slipher:
There are no translation markers in the engine. They were removed in
2014


illwieckz (approved):
LGTM;

PR #65, opened 2021-10-09
rocket: unify emoticon class with what people do by hand

illwieckz:
Also unify the looking and don't duplicate what's inherited.

Minor thing, maybe we want to delay the merge to the last minute because it needs a change on `Unvanquished` side.

See https://github.com/Unvanquished/Unvanquished/pull/1527


bmorel:
LGTM

PR #70, opened 2021-10-25
medkit usage belongs to BT

bmorel:


bmorel:
Note that this PR should be merged before the sister PR in master, and that it won't bring any problem if merged.
So, is there is any reason to not merge it?

bots/default.bt
```diff
@@ -9,6 +9,10 @@ selector
 		}
 	}
 
+	condition ( percentHealth( E_SELF ) <= 0.5 && haveUpgrade( UP_MEDKIT ) )
+	{
+		action activateUpgrade( UP_MEDKIT )
+	}
```
necessarily-equal, line 15: Untested, but this should fix the issue mentioned in the sister PR

```suggestion
	// Force this branch to be considered failed so that this doesn't stop the bot
	// from doing what is currently doing (for example fighting)
	decorator return( STATUS_FAILURE )
	{
		condition ( percentHealth( E_SELF ) <= 0.5 && haveUpgrade( UP_MEDKIT ) )
		{
			action activateUpgrade( UP_MEDKIT )
		}
	}
```

necessarily-equal (changes requested):

bots/default.bt
bmorel: which issue? Has it been seen in practice, or is it just some guessing from reading undocumented code?

bmorel:
superseeded by #1558

PR #72, opened 2021-11-01
Bring back bot behavior cvars

necessarily-equal:
This allows one to request bots to do, or avoid doing some behaviours

To be merged after https://github.com/Unvanquished/Unvanquished/pull/1576.

### DIFF
--- a/bots/default.bt
+++ b/bots/default.bt
@@ -38,7 +38,9 @@ selector
 					}
 					sequence
 					{
-						condition ( healScore <= 0.5 && percentHealth( E_GOAL ) > 0.3  && random > 0.3 )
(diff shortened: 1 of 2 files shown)


bmorel:
To me, that's a technically valid thing, but politically I'm not fond of your PR for few main reasons:

1. I don't like the idiom of splitting conditions one by one just to put them in a sequence: this makes the default behavior harder to read and hack, especially considering the syntax problems the current parser have (I won't talk about those on github, it requires a proper talk medium). It only contributes at helping making the syntax more obscure because all examples seems so complex...
2. I can't think about any "by practice" good effect.

PR #73, opened 2021-11-06
use new healScore function

bmorel:


necessarily-equal:
I was going to comment on how this didn't change the numbers while it probably should have been less easy to stop fighting to heal than to go back on healing when idling… except this does that already.

I guess the condition was backward for long.

PR #74, opened 2021-11-06
Medkit in BT & code move

necessarily-equal:


bots/unstick.bt
```diff
@@ -1,5 +1,16 @@
 // Subroutine to try to prevent bots from getting stuck
 
+decorator return( STATUS_FAILURE )
```
bmorel, line 3: should not be in unstick.bt imo, considering the fact this is human specific, and unstick applies to all

bots/default.bt
```diff
@@ -2,6 +2,8 @@ selector
 {
 	behavior unstick
 
+	behavior use_medkit
```
bmorel, line 5: good idea, makes the stuff more readable.

bots/unstick.bt
necessarily-equal: well, I can add an extra condition around it if you want, but I don't think it's a problem as the code would trigger in the exact same situations

bmorel:
good enough for me

bots/unstick.bt
necessarily-equal: note that we would probably want a similar thing for when humans and aliens are healing, and it would make a lot of sense to put these here, too

bmorel: that's not the problem. The problem I see is that it makes species-specific code into a generic file.

bmorel: about the healing, yes and no. To be honest, I don't remember how I do it in my betterai branch, but in the end should not care too much: the BT must be rewritten anyway, to be much clearer.

slipher:
Does the new unstick.bt even parse? I thought there has to be one top-level node.

bmorel:
That simple question questions more the proces  to me

PR #75, opened 2021-11-07
Fixup previous PR (caught by slipher)

necessarily-equal:


necessarily-equal:
This basically wraps the whole file in a `selector` tag. Merging as it fixes the bug.

bots/unstick.bt
```diff
@@ -1,48 +1,50 @@
...
+				action moveInDir(MOVE_RIGHT)
+			}
+			condition(stuckTime > 12500)
+			{
+				action moveInDir(MOVE_BACKWARD)
+			}
+			sequence
+			{
+				action moveInDir(MOVE_LEFT)
+				action moveTo(E_SELF) // cancel old target which may be causing the bot to walk into a wall, etc.
```
slipher, line 46: moveInDir always returns success so this line is unreachable

slipher: Oops, never mind. It's sequence not selector so that's OK.

PR #76, opened 2021-11-08
split default.bt to have one per faction

bmorel:
Splits the BT in 2, one per faction, and then apply some code clean up.
Previous number of lines was 103, next is 118, but much easier to maintain and improve.

Still, this requires some real testing, since experience shown that bugs can happen fast. This has been sent on my servers for that intent.

### DIFF
--- a/bots/default.bt
+++ b/bots/default.bt
@@ -2,102 +2,13 @@ selector
 {
 	behavior unstick
 
-	behavior use_medkit
-
-	sequence
+	condition ( team == TEAM_ALIENS )
 	{
-		condition alertedToEnemy
-		selector
-		{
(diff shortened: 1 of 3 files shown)


bmorel:
I won't merge.

necessarily-equal:
I merged the first commit to master. The other one needs more testing IMO, to avoid the bug where bots were acting weird (e.g. 7efa3ebe76aef27040e10133dceae5e7b2bbe42a, and there's an issue for that somewhere)

necessarily-equal:
I confirmed the second commit to make bots really, really weird / stupid. I could stand for a minute as an advanced dragoon atop the armory on plat23 before eventually getting killed.

So yeah, condition nodes really break STATUS_RUNNING actions.

PR #77, opened 2021-12-04
implement ghost-yellow shader

bmorel:
Those files were previously in res-buildables_src repo, but I think this
does not really makes sense.


illwieckz:
Just replace `yellowbuild.png` with optimized https://dl.illwieckz.net/j/yellowbuild.png and everything would be fine for me.

bmorel:
@illwieckz : done.

necessarily-equal (approved):
LGTM

slipher:
Links! https://github.com/Unvanquished/Unvanquished/pull/1617 https://github.com/UnvanquishedAssets/res-buildables_src.dpkdir/pull/7

I suppose it's smart to move it here, so that we can use the new feature without having to load more unreleased paks besides `unvanquished_src`. :+1:

PR #79, opened 2021-12-18
Removes hard-coded extensions from rml

bmorel:
Those extensions are not useful per se, as if they are absent the engine will search for a list of extensions instead.
Removing the extensions makes things less surprising, preventing one to search for an exact filename which might or not exist.

### DIFF
--- a/ui/hud_human.rml
+++ b/ui/hud_human.rml
@@ -260,17 +260,17 @@
 
 
 		<div id="hudbg">
-			<img id="right" src="/ui/assets/human/rightbg.tga"/>
-			<img id="left" src="/ui/assets/human/leftbg.tga"/>
+			<img id="right" src="/ui/assets/human/rightbg"/>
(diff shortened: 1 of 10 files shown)


necessarily-equal (approved):

necessarily-equal:
I haven't tested, but looks fine from a code point of view

ui/hud_human.rml
```diff
@@ -70,7 +70,7 @@
 				color:#aaffffff;
 			}
 
-			div.ammo, div.stamina{
```
illwieckz, line 73: This one is not extension related. I'm OK with it I just want to be sure it's intentional.

bmorel: This commit should more likely go in another PR actually, since indeed, it's not extension related

illwieckz: Ah sorry, I haven't noticed it was a different commit. It's fine then.

PR #81, opened 2021-12-30
Grenade use

bmorel:
Teach bots how to use grenades

### DIFF
--- a/bots/default_humans.bt
+++ b/bots/default_humans.bt
@@ -42,14 +42,9 @@ selector
 				condition ( distanceTo( E_FRIENDLYBUILDING ) > 400 )
 				selector
 				{
-					condition ( haveUpgrade( UP_FIREBOMB ) )
-					{
-						action activateUpgrade( UP_FIREBOMB )
-					}
-					condition ( haveUpgrade( UP_GRENADE ) )
-					{
-						action activateUpgrade( UP_GRENADE )
-					}
+					// no need to check the upgrade is in inventory, since action would fail
+					action activateUpgrade( UP_FIREBOMB )
+					action activateUpgrade( UP_GRENADE )
 				}
 			}
 			action fight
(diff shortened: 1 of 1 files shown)


slipher:
Can you make it distance to *any* human buildable, not just reactor? It's very frustrating when bots on my team destroy our own base (as they do with luci).

illwieckz:
Example of bot nading an outpost (video from September, on bot server):

[bot outpost grenade](https://dl.illwieckz.net/b/unvanquished/bugs/bot-grenade-outpost/20210912-032602-000.unvanquished-bot-grenade-outpost.webm)

In fact every wave of bots returning to armoury (bought and) dropped between 1 to 3 grenades in the base.

bmorel:
> Can you make it distance to any human buildable, not just reactor? It's very frustrating when bots on my team destroy our own base (as they do with luci).

Sadly, unvanquished lacks such feature, currently it would require listing *all* the buildings one by one. That feature would also be useful for aliens to defend properly when close enough to *any* alien building, and would ofc also be useful for lucifer.
Keep in mind that, currently, grenade buying is disabled by default, though, and that adding that "friendly" or "enemy" keywords in BT would require 2 PRs to be synced.

bmorel:
> In fact every wave of bots returning to armoury (bought and) dropped between 1 to 3 grenades in the base.

Interesting. Were they alerted to some alien presence?

slipher:
You could add a pseudo-entity for closest building of a team. Similar to how there is `E_ENEMY` for some enemy.

bmorel:
Yet, that'd be the solution, and it would be useful for much more AI stuff. Now, that E_ENEMY thingy is, AFAIK, not tested on the field, and adding E_ALLY (and E_NEUTRAL in the process, maybe? Which would exclude spectators, ofc) might be more work than expected.
Not really in the scope of this fairly simple PR.

slipher:
How about something like this https://github.com/Unvanquished/Unvanquished/pull/1666

slipher:
> How about something like this Unvanquished/Unvanquished#1666
merged

bmorel:
Nice, this will trigger few tweaks in the BTs, but let's start with this obvious one!

bmorel:
@slipher @illwieckz is that ok now? (considering the fact bots still can't buy grenades by default, which would require a PR to unvanquished itself to change)

slipher:
LGTM

PR #82, opened 2021-12-30
Evolve before fight

bmorel:
evolve before fight, only if enough hp

### DIFF
--- a/bots/default_aliens.bt
+++ b/bots/default_aliens.bt
@@ -1,6 +1,6 @@
 selector
 {
-	condition ( aliveTime > 1500 )
+	condition ( aliveTime > 1500 && healScore < 0.5 )
 	{
 		action evolve
 	}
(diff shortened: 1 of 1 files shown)


bots/default_aliens.bt
```diff
@@ -1,6 +1,6 @@
 selector
 {
-	condition ( aliveTime > 1500 )
+	condition ( aliveTime > 1500 && healScore < 0.5 )
```
slipher, line 3: healScore involves distance to friendly structures as well as HP. Is that intentional?

bmorel: It is, but yeah, it's arguable, same for the score I used.

bmorel: @Slipher: I could change it so that it depends on health, instead.

slipher: Just curious, fine with me either way

necessarily-equal:
This should help base defense as this lets bots evolve when close to enemies, but not in line of sight. Thanks!

bmorel:
Yep, the BT changes I propose are all based on ideas I got while observing bots endlessly :)

PR #84, opened 2021-12-30
prefer pulse to luci to defend base

bmorel:
buy equipment before going to repair, to have armors in case of attack

### DIFF
--- a/bots/default_humans.bt
+++ b/bots/default_humans.bt
@@ -7,6 +7,13 @@ selector
 		condition alertedToEnemy
 		selector
 		{
+			decorator timer ( 1000 )
+			{
+				condition haveWeapon( WP_LUCIFER_CANNON ) && distanceTo( E_H_REACTOR ) <= 400
+				{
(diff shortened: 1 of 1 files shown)


slipher:
I wonder if pulse rifle could be problematic with teamkilling the base too... Well, it's probably an improvement on the status quo, LGTM

bmorel:
> Well, it's probably an improvement on the status quo, LGTM

From my testing, it's globally better, but to be honest, it's far from perfect: distance might require tweaking, and distance of the target compared to any friendly building should also be taken into consideration. Sadly, those changes are a less trivial, and if using any building in current state of things, the code would be become less maitainable (and BTs are already not the friendliest piece of code)

bmorel:
Regarding pulse for friendly fire, it is true that sometimes they miss target, but pulse is a lot less dangerous than luci, which they *always* charge to 50% and which have a rather wide explosion radius

bmorel:
Nobody seems against.

PR #85, opened 2021-12-30
humans defend base instead of healing

bmorel:
instead of healing or waiting for medistat to be free, of course

### DIFF
--- a/bots/default_humans.bt
+++ b/bots/default_humans.bt
@@ -27,7 +27,9 @@ selector
 			{
 				selector
 				{
-					condition ( ( distanceTo( E_H_REACTOR ) <= 300 || distanceTo( E_H_MEDISTAT ) <= 300 ) && isVisible( E_ENEMY ) )
+					// defending reactor, or try to avoid being killed while healing (hence the different values)
+					// the isVisible check is here to prevent "defending" when enemies are detected by radars
+					condition ( ( distanceTo( E_H_REACTOR ) <= 300 || distanceTo( E_H_MEDISTAT ) <= 250 ) && isVisible( E_ENEMY ) )
 					{
 						action fight
 					}
--- a/bots/default_humans.bt
+++ b/bots/default_humans.bt
@@ -27,6 +27,11 @@ selector
 			{
 				selector
 				{
+					condition ( ( distanceTo( E_H_REACTOR ) <= 300 || distanceTo( E_H_MEDISTAT ) <= 300 ) && isVisible( E_ENEMY ) )
+					{
+						action fight
+					}
+
 					sequence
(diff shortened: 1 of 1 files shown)


bots/default_humans.bt
```diff
@@ -27,6 +27,11 @@ selector
 			{
 				selector
 				{
+					condition ( ( distanceTo( E_H_REACTOR ) <= 300 || distanceTo( E_H_MEDISTAT ) <= 300 ) && isVisible( E_ENEMY ) )
```
slipher, line 30: What's the idea of checking distance to medistation here?

bmorel: yep, this is one of the things which can be improved with it

bmorel: I'm thinking, though, is it really desirable to make hurt humans defend *any* human settlement?
Reactor and medistat are rather obvious: they are *needed* to be able to heal, but defending defenses instead of using them as cover... And this will apply to aliens, too (which currently only defend the overmind, not boosters!)

bmorel: @slipher : what's your opinion? Should it be merged as it is, should bots help defenses even if wonded, or any other idea?

slipher: Maybe only the reactor is fine. The medistation in particular seemed a strange choice, since it would make bots less likely to heal close to a medistation.

bmorel: The point with medistat is that if enemies are close to medistat, they might kill the bot when the bot is healing. Of course, due to the `isVisible` check, alien need to attack the bot from front to, in theory (I never tested, the timing is quite hard), be attacked, but if at some point the heal action makes bots turn on themselves to seek for enemy, this theoretical improvement should show (or fail, ofc).

slipher: So the two conditions have different purposes? Like the reactor one is to defend base, and the medistation one is to avoid standing passively on the medistation while attacked? Maybe the medistation condition should use a shorter distance in that case.

BTW, you know you can write comments in BTs right? Same styles as C++.

bmorel: >  Maybe the medistation condition should use a shorter distance in that case.

Good point.

> BTW, you know you can write comments in BTs right? Same styles as C++.

I know, but I never think about doing it :)

bmorel: @slipher : I added a comment and changed the value slightly, imo, as for all hard coded values in BT, this will very likely require calibration, obviously (or suggested values, I'm ok to adopt any, it's not *that* important at this point). Are there other points to address?

slipher: LGTM

PR #86, opened 2021-12-30
humans roam near medistat if medistat is busy

bmorel:
Note that this require «action heal» to be able to fail to work.


slipher:
Looks like this depends on https://github.com/Unvanquished/Unvanquished/pull/1598

bmorel:
It does, but I see no reason to not merge it. I have pinged @afontain several times on it, and am very close to just merge it myself.
If this PR goes approved, I'll just do it.

bmorel:
@necessarily-equal said "LGTM" on irc.

PR #87, opened 2022-01-04
upscale inventory icons

bmorel:
Passes the size of humans inventory icons from 1.5em to 1.8em.
I do not know if those are optimal values, but I think this improves the situation. Icons still needs improvements imo, though.

Comparison:

2022-01-04-151415_1366x758_scrot
2022-01-04-151448_1366x758_scrot


slipher:
LGTM

PR #88, opened 2022-01-04
Tab indent

bmorel:
Don't mix tabs and spaces in ui's stuff

### DIFF
--- a/ui/human_buy.rml
+++ b/ui/human_buy.rml
@@ -13,11 +13,11 @@
 			}
 
 .circlemenu icon{
-  display:inline-block;
-  position:absolute;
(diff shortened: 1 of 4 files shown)


bmorel:
Note that I purposely *not* changed the different styles used for blocks (have the '{' on same line or not), because:

1. it's *a lot* less annoying and
2. it's *a lot* more mixed (thus can't guess what it should be in less than 30s)

illwieckz (approved):
LGTM. 😉

slipher:
I was working on resurrecting the RmlUi branch a couple weeks ago and it's almost finished.

Don't do it or I'll have to start over :sob:

PR #89, opened 2022-01-12
use resupply BT API

bmorel:
Depends on an unvanquished PR: "Ai resupply #1698"

### DIFF
--- a/bots/default_aliens.bt
+++ b/bots/default_aliens.bt
@@ -32,6 +32,12 @@ selector
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


bmorel:
I think the score values should be tweaked, but that's a rather boring task to do: tweak, observe bots, and especially humans for ages, re-tweak, etc...
I'm planning to completely re-implement the "/give" command to reduce the bot testing pain as much as possible, or maybe to "fork" it for the need, to prevent breaking compat (on a cheat command? Not sure that'd be much useful...).

bmorel:
Updated so that reloading action no longer prevents acting.

PR #91, opened 2022-01-13
add UI for radarRatio

bmorel:


bmorel:
This breaks the GUI, for some reason.

PR #95, opened 2022-01-17
Migrate from libRocket to RmlUi

slipher:
Companion https://github.com/Unvanquished/Unvanquished/pull/1715

### DIFF
--- a/ui/alien_build.rml
+++ b/ui/alien_build.rml
@@ -19,15 +19,15 @@ body
 	font-size: 1.7em;
 }
 .circlemenu button {
-	icon-color-multiplier: red;
+	image-color: red;
 }
 .circlemenu button.locked{
-	icon-image: /ui/assets/circlemenu/circle-hover;
-	icon-color-multiplier: rgb(50,0,0);
+	decorator: image( /ui/assets/circlemenu/circle-hover );
+	image-color: rgb(50,0,0);
 }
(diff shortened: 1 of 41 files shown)


slipher:
Event names MUST be lowercase in RmlUi.

Size tweaks were needed because the CSS engine changed or something, so that some elements ended up with different sizes with RmlUi vs. with libRocket.

illwieckz:
The RML part looks good to me (as soon as it produces the expected result).

bmorel:
> Event names MUST be lowercase in RmlUi.

Ah, sad.
But looking at libRocket's examples, it is likely that this could work with libRocket too, so it would be possible to get that part in a different PR for immediate merge, which is easier to try, especially now that there is a nightly build server. It would also make the review significantly easier (even if I already looked at the code and seen no problem).

illwieckz:
I added a commit to fix the server browser tabs to be displayed on the left. This is the only visual discrepancy I have seen so far.

illwieckz:
Oh no, this one was merged to `master`.

illwieckz:
I merged the branch to `0.53.0/sync` instead and returned `master` to initial state.

illwieckz:
if you have pulled unvanquished_src.dpkdir master before, you have to reset to 2b70d5cb3819559faa73d228c12fe1fd54b9de9f

PR #96, opened 2022-01-18
reduce unstick times

bmorel:
This is something I had in previous betterai's BT, that I lost when I wanted to cleanup everything. Interestingly enough, a player noticed very quickly the change of behaviors.
Those values are like the ones I used before, more or less, that is, mostly divide times by 2. Other values could be used I guess, but one must keep in mind that unvanquihed is a fast game, and 10s doing nothing is a serious problem.

### DIFF
--- a/bots/unstick.bt
+++ b/bots/unstick.bt
@@ -12,7 +12,7 @@ selector
 	}
 
 
-	condition(stuckTime > 10000)
+	condition(stuckTime > 5000)
 	{
 		selector
 		{
@@ -21,22 +21,22 @@ selector
 				action suicide
 			}
 
-			decorator timer(3000)
+			decorator timer(1500)
 			{
 				decorator return(STATUS_FAILURE)
 				{
 					action jump
 				}
 			}
-			condition(stuckTime > 17500)
+			condition(stuckTime > 9000)
 			{
 				action moveInDir(MOVE_FORWARD)
 			}
-			condition(stuckTime > 15000)
+			condition(stuckTime > 7000)
 			{
 				action moveInDir(MOVE_RIGHT)
 			}
-			condition(stuckTime > 12500)
+			condition(stuckTime > 6000)
 			{
 				action moveInDir(MOVE_BACKWARD)
 			}


slipher:
Now the bot will try to run in the same direction for the last 11 seconds of the sequence. The 20 seconds number could be reduced, but of course this was chosen to avoid "cheating" by having a suicide timer longer than human players can.

Running out of stamina may be a concern with so many jumps.

bmorel:
> Now the bot will try to run in the same direction for the last 11 seconds of the sequence. 

Indeed, but end result is still better than 

> The 20 seconds number could be reduced, but of course this was chosen to avoid "cheating" by having a suicide timer longer than human players can.

You mean smaller. Well, you're right, but note that I have not touched that value commanding suicide.
TBH I didn't though about that, but the only way I see to have this reliable would be to have bots obeying the same rules as players, enabling the `/kill` command when they are stuck, and disabling it when they reached a proper place.

> Running out of stamina may be a concern with so many jumps.

It can be, and it could be fixed rather easily in the C++ side. Or it will require a stamina condition (and access to stamina cost of actions, too), which *also* requires C++ changes (it might still be desirable though).

The unstick behavior was a workaround, and still is. What this PR try to do (and in my experience it helps) is improve on it by reducing times: there are no more marauders idling on their crate for 10s (which is, again, *a lot* in unvanquished).
There are still various problems with the pathfinding, notably when bots fall of their navmesh on a disconnected navmesh, because they try to reach goal despite it being unreachable, but this can't be fixed solely with a BT change.

bmorel:
> Running out of stamina may be a concern with so many jumps.

Actually, it is not, now that I think about it.
If they are stuck, then they are likely not alerted(*) and thus they are walking, recovering their stamina. Maybe the dance could be reworked to alternate more the different actions, it would likely give better results in many situations.

*: although technically for that, we'd need to put the unstick dance at the end of the BT, not at start, or at least after the `alertedToEnemy` stance, which would even fix some of the problems I have with it... I'll give that idea a try!

PR #97, opened 2022-01-29
add a decorator near the heal actions

bmorel:


slipher:
The decorator needs to be around the `heal` action, since that is the one that can rapidly toggle between running/failure if the bot is at distance 200.

The enclosing selector on the other hand, shouldn't ever fail since roaming should always work.

slipher:
LGTM

PR #98, opened 2022-02-08
remove no longer used option cg_depthSortParticles

### LINKED ISSUE #20: remove option for cg_depthSortParticles in UI
Enabling `cg_bounceParticles` while disabling `cg_depthSortParticles` is wrong and produce issue users will report as bugs (like missing snow in chasm, or buildable remains not laying on floor).

Either `cg_bounceParticles` is enabled and does depth sort, either it's disabled. We may even remove the `cg_bounceParticles` cvar, but at least, we must remove the “Ensure over-lapping particles are rendered correctly” option from the UI.

There is no reason on earth to use `cg_bounceParticles` whithout `cg_depthSortParticles`.

bmorel:
fix #20

### DIFF
--- a/presets/graphics/high.cfg
+++ b/presets/graphics/high.cfg
@@ -20,6 +20,5 @@ cg_shadows 1
 cg_playerShadows 1
 cg_buildableShadows 0
 cg_bounceParticles 1
-cg_depthSortParticles 1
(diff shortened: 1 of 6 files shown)


illwieckz (approved):
LGTM

PR #99, opened 2022-02-13
New icons

Bob-Vador:
New icons, what do you think?
What do you think about the black line and the armor icons?


Gireen:
The icon images are supposed to be only white, so no black lines please. 

The armor icons are to detailed and hard to distinguish, especially light and medium armor.
I think something more symbolic like only the small badges but more clear would be better.

Bob-Vador:
Capture d’écran_2022-02-14_13-42-51
That look like that in game.

bmorel:
> The icon images are supposed to be only white

What is the technical reason behind it? If it is only for aesthetics, I don't see why it can not change, especially since it's already broken (by the grenade icons, that I do not like neither mind you).

> The armor icons are to detailed and hard to distinguish, especially light and medium armor

To give a base to talk for all, 2 screenshots, with and without, in front of same armoury (plat23, one can ignore the double icon thing, it's a bug and unrelated to this discussion):

2022-02-16-214435_1366x768_scrot
2022-02-16-214738_1366x768_scrot

To be fair, I agree that the light/medium armors are worst, since it's harder to distinguish between them, but for BS it's better than original imo.
For the other ones, which takes part in the HUD, the outline is thin enough to be barely noticeable, yet it makes things easier to see globally.

necessarily-equal:
I really like this. I would make the light armor shield smaller so that it is possible to distinguish them on first sight

Gireen:
> What is the technical reason behind it? 

Its more flexible to use, easier to identify and scales better

Maybe you forgot the discussion you and i had in the forum a few day ago about this exact topic.
https://forums.unvanquished.net/viewtopic.php?f=8&p=18691

to quote myself:

> It becomes unnecessary complicated to change. Best example is try to change the icon color to something dark like black. It will turn into a blob so we would require extra versions for every icon if they had "hardcoded" outlines.
>
>An outline should be added, if at all, in the game it
self where it would necessary and not in the graphics.
>
>To be clear i think it is a good idea to vectorize all icons, cause i would like to have them turned into an icon font.
>But adding black outline to all images makes future changes harder not easier and creates unnecessary work.

Just look at the momentum bar. The armor icons are an ugly mush cause of the black lines, while the white icons are clearly visible. 

>If it is only for aesthetics, I don't see why it can not change, especially since it's already broken (by the grenade icons, that I do not like neither mind you).

Two wrongs don't make a right. As you might also forgotten we had a forums topic about this exact issue with "over detailed icons".
Nanaa was so nice and created new grenade icons.
d1hMt2p

bmorel:
> Nanaa was so nice and created new grenade icons.

I agree that those icons are nice, and I would love to have those icons merged, too.

> while the white icons are clearly visible.

I think this is because of the number of thin details, not because it's not only transparent and white. This applies only to armors icon in that PR, and I agree with you on that one.

illwieckz:
@Gireen do you know where we can find those icons by Nanaa?

Gireen:
@illwieckz https://forums.unvanquished.net/viewtopic.php?p=18488#p18488

illwieckz:
Ah thanks, I also see I already downloaded them, nice. Well those two nades icon are very good, let's import them !

Bob-Vador:
@Gireen, what do you think about put only the shield in the momentum bar?

bmorel:
ping @Gireen

Gireen:
I think that was covered in an IRC discussion at the time.
also 

>  I think something more symbolic like only the small badges but more clear would be better.

Different sets of icons make it only more confusing. 

That's also my last input on this topic. My interest in this unnecessary lengthy and redundant discussions is exhausted. Feel free to do what you want and add any cryptic icons with as much outlines as your heart contents and physical possible. 😂

sweet235:
I guess this was neither rejected nor accepted, but simply forgot. Why is that?

I think the icons are fine, except that the difference between light and medium armour could be more obvious.

Bob-Vador:
These icons have a black outline and the icons must have no outlines to be accepted (I'm working on that)

DolceTriade:
cool, gireen gave me his icons: https://libera.ems.host/_matrix/media/r0/download/matrix.org/iqJMNGSopClsWLGZsUDAqzKm/building-icons_0.1.dpk

Also I stalked Bob-Vador and found his: https://github.com/Bob-Vador/unvanquished-icons

We should mix and match these.

DolceTriade:
Hey Bob, these look good. The one feedback I'd have is if we could have more differentiation between the light armor and medium armor icons. The extra lines aren't super obvious.

It looks like this now:

bmorel:
Has to be noted that currently, the difference between light and medium armour is NOT OBVIOUS at all. It's just a letter change, after all, so I'm fine with merging this as it is, already.
If more improvements can be done before merging, all the better, ofc, but I don't think it's a stopper.

DolceTriade:
Fair. I'm fine to merge these in their current state and get feedback and iterate from there.

Bob-Vador:
I know but it's difficult, I can modify a bit the light armor but I have no other idea. I think I going to keep the extra lines and if someone have an idea, let me know.

DolceTriade:
It's probably fine. Worst case, we'll put the L and M back on there :weary:

illwieckz:
My concern with those icons is that they look good when looked at separately, but I think they don't improve reading when put together. I first seen them appealing and I like the look of them but after having tested them in game my opinion changed.

Purpose of symbols is to identify clearly and quickly what things are about. That's why a stylized armor, even if it doesn't look like the one in game, may be efficient at doing it. That's also why even a shield symbol, even if there is no such item in game, can do the work.

Same apply for letters, there are many problems in using letters, but those are symbols very easy to discriminate. Problems with letters is that they are culture and language centric, but even without knowing the meaning of L and M in their original language or even without knowing the latin alphabet, letters are meant to be easy to discriminate (that's why they even exist, those symbols inherit thousands of years of selections, design and polishing to be efficient).

Here the proposed icons, while looking good, are more depicting the in game model than carrying a meaning. That even assumes the in game model would be good enough to carry the meaning to begin with, and that a derivation removing some strong semantic carrier like color would keep the ability to carry the meaning.

To me the simple fact there was a need to add a second symbol (a shield) to carry the meaning shows there is a problem. It's like using two words when it would be better to use one, like using a periphrasis to ask for help instead of using the “help” word.

I'm not comfortable with this because in a way I like the looking of those icons, but for example the medium armor symbol look like to me a more detailed light armor symbol, not a symbol for a more-protective armor. And a single icon should be enough (either armor, either shield).

The fact we use the same model for light and medium armor doesn't help of course when the icon design process is meant to derivate from the in game model, but how much symbols should derivate from model (or how less the should?)

illwieckz:
One good thing though: at first I also found the black outline nice, but after testing them in game this changed my mind as well. So it's a good thing they were remade without black outlines. We better want to merge icons without outlines whatever they are, so that is welcome.

illwieckz:
About making things easy and quick to read, why not use a common armor base for both medium and light but display the light one with less elements?

Here are a some quick mockups reusing our current armor icon (quick edit in GIMP), just to illustrate the idea:

[armor mockups](https://dl.illwieckz.net/b/unvanquished/bugs/circle-menu-icons/unvanquished_2022-09-19_134726_000.jpg)

[armor mockups](https://dl.illwieckz.net/b/unvanquished/bugs/circle-menu-icons/unvanquished_2022-09-19_143840_000.jpg)

I'm not saying we should keep that old icon, but the idea is that we may have a stylized medium armour that looks more bulky and covering more parts than light variant.

Readability and very fast readability is a very strong need. One can hardly make more obvious than a letter change (that's probably one of the most trained algorithm in human brain after face recognition) but using a letter in such context is just using a placeholder, so that looks like a placeholder (= bad). Like most word should be figurable out of 3 or 4 syllables, only a handful of shapes could be enough to carry the meaning and makes things easily and quickly discernible.

Bob-Vador:
It's what I want to do with my icons for finally remove the extra lines. The shields Can be removed but used in the momentum bar.
I also think that in all cases the bsuit icon need to be reworked

illwieckz:
It's better if we can use the same icon in circle menu and momentum bar (it would be confusing otherwise). I believe everyone agrees all armours can get better drawings indeed

bmorel:
I think @illwieckz 's idea of "why not use a common armor base for both medium and light but display the light one with less elements" is good.
This is, indeed, what @Bob-Vador  tried to do, but with detail added for medium. Maybe this is where the problem lies. Which includes the fact the model difference between light and medium does NOT EXISTS AT ALL (for reminder).

Merging the 2 ideas might lead to something like using @Bob-Vador 's light armour icon, but with empty (transparent, with a white "outline") areas between legs|arms and chest ?

Bob-Vador:
Yeah, I think it's a good idea, thank you, I will try to do some thing like that this weekend, it can be great I think. Thank you !

PR #102, opened 2022-02-21
UI improved start local

bmorel:
This PR aims at improving the "Start Local/LAN Game" GUI, notably by allowing to enable `/devmap`, editing the CVar `fs_extrapaks` and specifying a map layout.
Ideally:

* the list of maps would changed for a radiobox-list
* the available map layouts would be listed
* a placeholder would be specified instead of the levelshot when no map is selected yet, to avoid the annoying auto-resize

Currently, the layout selection code does not work (hence the draft status). The devmap and the extrapaks are fine though, while they very probably could use style improvements, but I'll let CSS experts handle that part I think, my priority is to make it easier for people to contribute to maps.

### DIFF
--- a/ui/server_create.rml
+++ b/ui/server_create.rml
@@ -19,7 +19,7 @@
 	</head>
 	<body template="window" id="createserver" onshow='Events.pushevent("buildDS mapList", event)' style="width: 40em; margin: 1em;">
 		<h1><translate>Start local/LAN game</translate></h1>
-		<form onsubmit='Events.pushevent("execForm \"map $map$\"", event)'>
+		<form onsubmit='Events.pushevent("execForm \"map $map$ $layout$\"", event)'>
 			<row>
 				<input type="text" cvar="sv_hostname" class="text" id="hostname"/>
 				<h3><translate>Hostname</translate></h3>
@@ -33,6 +33,11 @@
 			<row>
 				<input type="text" cvar="fs_extrapaks" class="text" id="extrapaks"/>
 				<h3><translate>Extra packages to load</translate></h3>
+			</row>
+
+			<row>
+				<input type="text" class="text" id="layout" name="layout"/>
+				<h3><translate>Map building layout to use</translate></h3>
 			</row>
 
 			<row id="maplist">
--- a/ui/server_create.rml
+++ b/ui/server_create.rml
@@ -38,12 +38,12 @@
 
 			<row>
 				<input type="text" cvar="g_password" class="text" id="password"/>
-				<h3><translate>Password </translate></h3>
+				<h3><translate>Password</translate></h3>
 			</row>
 
 			<row>
 				<input type="checkbox" id="devmode" style="margin-right:1em;"/>
-				<h3><translate>Developper mode (enable cheats)</translate></h3>
+				<h3><translate>Developer mode (enable cheats)</translate></h3>
 			</row>
 
 			<row>
--- a/ui/server_create.rml
+++ b/ui/server_create.rml
@@ -19,24 +19,29 @@
 	</head>
 	<body template="window" id="createserver" onshow='Events.pushevent("buildDS mapList", event)' style="width: 40em; margin: 1em;">
 		<h1><translate>Start local/LAN game</translate></h1>
-		<row>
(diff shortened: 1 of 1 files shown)


bmorel:
Reopened the PR and ported it for 0.53.1.
Before any other thing: the lack of map name showing in the dropdown list is a regression currently present in master, and the layout selection actually works.

Here are screenshots of what it currently gives (no map selected, map with minimap selected, map without minimap selected):

2022-08-23-203014_1366x768_scrot
2022-08-23-203022_1366x768_scrot
2022-08-23-203032_1366x768_scrot

ui/server_create.rml
```diff
@@ -19,24 +19,29 @@
...
-
 		<form onsubmit='Events.pushevent("execForm \"map $map$\"", event)'>
+			<row>
+				<input type="text" cvar="sv_hostname" class="text" id="hostname"/>
+				<h3><translate>Hostname</translate></h3>
+			</row>
+
+			<row>
+				<input type="text" cvar="g_password" class="text" id="password"/>
+				<h3><translate>Password </translate></h3>
```
necessarily-equal, line 41: There is a weird extra space here, but it's not a real problem

ui/server_create.rml
```diff
@@ -19,24 +19,29 @@
...
+			</row>
+
+			<row>
+				<input type="text" cvar="fs_extrapaks" class="text" id="extrapaks"/>
```
necessarily-equal, line 50: Stupid question maybe, but did you test this works? I would guess it works, but just from reading I am unsure because I remember there is some trouble where fs_extrapak must be passed with -set instead of +set because it would arrive too late otherwise.

ui/server_create.rml
```diff
@@ -16,10 +16,21 @@
...
+				end
+				return 'execForm "map $map$ $layout$"'
+			end
+		</script>
+
 	</head>
 	<body template="window" id="createserver" onshow='Events.pushevent("buildDS mapList", event)' style="width: 40em; margin: 1em;">
 		<h1><translate>Start local/LAN game</translate></h1>
-		<form onsubmit='Events.pushevent("execForm \"map $map$ $layout$\"", event)'>
+		<form onsubmit='Events.pushevent(MapMode(document), event)'>
```
necessarily-equal, line 33: I wonder if we could extend g_persistDevmap or a second cvar to open the map as devmap. I feel like it probably wouldn't worth it?

necessarily-equal (approved):
LGTM, haven't tested. Some remarks attached

ui/server_create.rml
bmorel: nitpick is certainly nice to encourage contributions.

ui/server_create.rml
bmorel: Last time I tried, monthes ago, I think it worked.

ui/server_create.rml
bmorel: this PR was opened before the changes you're mentioning I think. Please take that kind of stuff into consideration, because it's quite frustrating to see things introduced after generating that kind of remarks.

bmorel:
You dropped comments, but still approved. What is your exact opinion here?

ui/server_create.rml
slipher: It's important to have clean strings for translators. Also if there were any existing translations, this would break them.

ui/server_create.rml
```diff
@@ -30,6 +41,11 @@
 				<h3><translate>Password </translate></h3>
 			</row>
 
+			<row>
+				<input type="checkbox" id="devmode" style="margin-right:1em;"/>
+				<h3><translate>Developper mode (enable cheats)</translate></h3>
```
slipher, line 46: Spelling: "Developer"

ui/server_create.rml
slipher: There's no trouble here.

ui/server_create.rml
necessarily-equal: note I'm not saying it has to be done, just wondering if it could

necessarily-equal (approved):
LGTM, haven't tested

PR #103, opened 2022-02-21
add an entry to (dis)allow bots to equip ckit

bmorel:


necessarily-equal (approved):
This works, can merge

I'd add a second check at the beginning of `BotActionRepair` though, this would avoid mistakes with the non-default BT. In the current state you can have a condition where bots loop trying to get a CKit to repair.

PR #107, opened 2022-03-17
rework flamer damage

illwieckz:
@bmorel

### DIFF
--- a/configs/missiles/flamer.attr.cfg
+++ b/configs/missiles/flamer.attr.cfg
@@ -1,5 +1,5 @@
 // attributes
-damage             10
+damage             2
 meansOfDeath       MOD_FLAMER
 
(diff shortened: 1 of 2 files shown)


configs/missiles/flamer.attr.cfg
```diff
@@ -1,5 +1,5 @@
 // attributes
-damage             10
+damage             2
```
slipher, line 2: This is a useless attribute - see #106.

illwieckz: This is `configs/missiles/flamer.attr.cfg`, not `configs/weapon/flamer.attr.cfg` (`missiles` instead of `weapon`).

illwieckz:
Let's do it, we can make it stronger if needed.

PR #109, opened 2022-05-28
rework flamer damage

illwieckz:
Increase the direct damage, people said it was too low.

I increased to `5`. Historically it was `10`, I then reduced it to `2`, people says it's not enough. So direct damage is currently half of what stock 0.52 did.

```
<afontain> also the flamer nerf is probably too strong
<freem> flamer is currently useless with the nerf
<freem> damage nerf should likely be removed, only the ignition nerf have a use
```

One of the big problem of the flamer outside of the ability to paint fire on the floor and almost immediate ignite buildables is that it had both high damage and high rate which made it a very powerful mid-range painsaw.

With `10` one would only have to fire one second to kill a tube. The range of the flamer is even larger than the range of a tube so with `5` a naked human can rip out all the tubes of the default plat23 base without being hurt if no alien defends. With `5`, a player carrying a bsuit can do the same while taking damages.

Thanks to the nightly server the previous attempt was tested, in the same way we can merge this and get this delivered to the nightly server.

### DIFF
--- a/configs/missiles/flamer.attr.cfg
+++ b/configs/missiles/flamer.attr.cfg
@@ -1,5 +1,5 @@
 // attributes
-damage             2
+damage             5
 meansOfDeath       MOD_FLAMER
 
 splashDamage       10


bmorel:
What if the range was decreased to be inferior to acid tube, and damages less nerfed? Say, 7 damages (that's already a 30% nerf) but to only 8 meters?

necessarily-equal:
So to recap, the summary of the nerf if we were to merge this in the current state would be

* direct damages divided by 2
* ignite chance divided by 2.5 (from 50% chance to 20%)
* ground fire chance divided by 3 (from 30% chance to 10%)

This still feels like a large nerf, with every stat being nerfed by more than 50%.

It seems however that the flame has some splash damage that wasn't changed, which ponders this a bit https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/blob/master/configs/missiles/flamer.attr.cfg#L5

This is probably still a huge nerf, and while I was playing with flamer for https://github.com/UnvanquishedAssets/res-weapons_src.dpkdir/pull/11, I noticed that the flame seemed very capricious (too capricious) to drop flames on the ground. I'd say this looks bad, so I'd revert course on that too.

Maybe if we want to make ground flames less dangerous we should lower their damage instead.

illwieckz:
> This still feels like a large nerf, with every stat being nerfed by more than 50%.

I remind you that ignited buildables continue to receive huge damage after one or three hits, even while the player initially carrying the flamer is long dead. This is basically why the flamer needs a huge nerf: even with 90% nerf on direct hit the flamer would still be a weapon to spawn chainsaws right inside buildables. The flamer is still very strong as an indirect weapon.

bmorel:
Then maybe the answer is to reduce the DoT damage and the ignition chances?
This would affect the firebomb too, which is currently considered as more interesting than the grenade by most players I know, while being cheaper.

illwieckz:
> I noticed that the flame seemed very capricious (too capricious) to drop flames on the ground.

This was intended by my initial patch: to avoid painting fire while running backward, which is overpowered.

Also, even if we don't have to reach heavy realism, the fact that the flamer was generating more flame when hitting the ground than when fired in air was super weird.

bmorel:
> Then maybe the answer is to reduce the DoT damage and the ignition chances?

And it would not affect direct damages of the flamer much, so it should feel less useless than with too big direct nerf

illwieckz:
> reduce the […] ignition chances?

This is because I reduced ignition chances it's now capricious to drop flames on the ground.

bmorel:
Yes. What I mean is, keep the reduced ignition chances (maybe reduce it less, I do not have a strong opinion on this matter) but reduce the damage over time instead of the direct damages.
Which would thus also affect the damage over time of the firebomb, which is currently considered better than the grenade by most players I know, while being cheaper.

I hope my point is clearer like this  :)

illwieckz:
I just remind a single man with one flamer (just because he carries a flamer) was able to defeat a whole team of aliens combining cleverly their efforts. Flamer was like a chainsaw spawning chainsaws within buildables and painting chainsaws on floor.

I'm not against a more complex mechanism like damage being reduced over time (nice idea, really!), but I don't want to postpone 0.53.0 for this, and I won't implement it. Right now I'm just aiming for a nerf that is implementable easily like this, and just prevents the flamer to kill the game like it does for now. We can do better later, better fix the huge imbalance now.

The best way to have an opinion on that direct damage change is to merge it anyway, so let's do it and see.

bmorel:
> I'm not against a more complex mechanism like damage being reduced over time (nice idea, really!),


Was not what I meant. I meant, fire on ground or in buildables in a "damage over time" thing, and it's raw value could have been lowered, leading to a nerf to all fire, being flamer itself or firebomb.

PR #111, opened 2022-06-01
Update unstick.bt

superdupont-fr:
Extented with a second part, from 3 to 5 seconds without suicide.
The first part, was modified a few, and apply for 7 to 15 second and end with suicide in case still stuck.

### DIFF
--- a/bots/unstick.bt
+++ b/bots/unstick.bt
@@ -1,4 +1,5 @@
 // Subroutine to try to prevent bots from getting stuck
+// There is 2 parts, one from 3 to 5 seconds, and the second part to 7 to 15 seconds.
 selector
 {
 	decorator return( STATUS_FAILURE )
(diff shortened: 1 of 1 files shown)


slipher:
I'm wary of triggering it after so little time (3 seconds) as this will frequently happen when the bot is not really stuck.

Can you give some qualitative explanation of how this improves the unsticking performance? Why is it good to move in a direction for only 100 ms and then change direction?

superdupont-fr:
> I'm wary of triggering it after so little time (3 seconds) as this will frequently happen when the bot is not really stuck.

In fact, I tried with time at 1 seconds to ensure quicker unstuck. Worked fine, but not easy to debug for the moment, so, I keep the 3 seconds time.

> 
> Can you give some qualitative explanation of how this improves the unsticking performance? Why is it good to move in a direction for only 100 ms and then change direction?

In many case, moving for short time like 100ms or 200ms in diferent directions before trying to roam can unstick in many case. 
Long sequence of walking in same direction is useless. I watched bots in Thunder stucked in large corridor ending to suicide because of long time in same direction.
Also, the mechanism that check if the bot is still stuck checking each seconds, that I think is too long. Many bots have time to come back at the almost same position in this time. My first BT with several small steps was already a good improvement.
My second update introduce a check if there is enemy nearby, and lets bots react if enemy appear in the sequence for unstick.
My third update is almost the same, but with include behavior to shorter files and improve readability. (I'm doing the third update now).

superdupont-fr:
The 4th update will be without debug info.

bots/unstick.bt
```diff
@@ -1,49 +1,226 @@
...
 				decorator return(STATUS_FAILURE)
 				{
-					action jump
+				sequence
+					{
+						condition distanceTo( E_ENEMY ) <= 300
+						{
+						sequence
+							{
+								action resetStuckTime
```
necessarily-equal, line 38:
```suggestion
						condition ( alertedToEnemy && distanceTo( E_ENEMY ) <= 300 )
						{
							sequence
							{
								action resetStuckTime
```
I think you are calling `resetStuckTime` accidentally if you don't check there is an enemy. Otherwise, if I'm reading correctly, the distance to E_ENEMY is 0 if there is no enemy in sight.

necessarily-equal (changes requested):

bots/unstick.bt
superdupont-fr: Always need alertedToEnemy with distanceTo( E_ENEMY ) ?

necessarily-equal: Yeah basically, otherwise you'll read a distance of 0

superdupont-fr: But, it's worked in my tests. (of course, without debug commands, I can't be sure that it's works really).

superdupont-fr: I more understood some BT parts. I rewrote some parts. Will test more before update.

illwieckz:

PR #114, opened 2022-06-09
different stamina icons when humans run or sprint

bmorel:
There is one known bug: when the human 1st spawn, both icons are shown. I have no idea why.
Sprint icon is courtesy of bob-vador, as said in the commit importing it.

This has been in production since several months on the betterai mod, on my servers, so you can judge by yourself if this helps or not, or if it makes things too bloated (I don't see how, since it only replaces one icon with another).

Also note that this should be perfectly compatible with rmlui. Put as draft on @necessarily-equal 's request.


bmorel:
FWIW, here is how it looks:

before:

old

After:

new-run
new-sprint

necessarily-equal:
replaced by https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/116 (merged)

PR #116, opened 2022-06-13
Add sprinting indication

necessarily-equal:
For now this is only a color. I prefer it without an icon, because I find that is harder to recognize than a color and thus I think it requires more mental effort to process it. It is also really hard to animate smoothly, which is why I'd rather avoid it. But depending on if @bmorel really wants an icon changing on sprint this is also possible.

Needs https://github.com/Unvanquished/Unvanquished/pull/1847.


illwieckz:
Maybe we can do both? Not everyone may process colors correctly. Color-based games like [Frozen Bubble](http://www.frozen-bubble.org/) also paint various symbols on colored marbles, and I guess it's on that purpose.

necessarily-equal:
I've added the icons, I still think it looks somewhat worse of with those and that they aren't needed, but well, here you go.

I managed to animate the icons though, so a large part of my complain about "it blinks" went away. That's good.

necessarily-equal:
Well, anyway here's some picture because it's important yada yada :P

image

image

necessarily-equal:
I should add that there are no know bugs in this PR, and that icons are courtesy of @Bob-Vador, with the borders removed because it didn't work well with antialiasing

PR #117, opened 2022-07-04
More luci protections

bmorel:
This increase the luci-base spamming protection, by:

1) including armouries in the list
2) increasing the distance from "sensible buildings"

Still not perfect, I know. I also have doubts about 2nd commit, tell me if you're rather drop it.


bmorel:
I removed the commit to increase range to 600, because I'm afraid it would introduce more problems than it would solve, keeping only the parts I consider safer before 0.53.
The way it makes things, area around RC and armouries should be a bit safer for humans.

necessarily-equal (approved):

PR #119, opened 2022-07-07
Human bot heal fix

bmorel:
At some point, the medkit condition was reinjected in the code (I had it removed for a long while in my mod). This prevents bots bot heal as long as they have the medkit, which means it's possible for a bot with 90% hp to walk on medistation and not stop to heal.
I took the occasion to reduce a big of ugliness.

### DIFF
--- a/bots/default_humans.bt
+++ b/bots/default_humans.bt
@@ -69,7 +69,6 @@ selector
 	{
 		sequence
 		{
-			condition !haveUpgrade( UP_MEDKIT )
 			condition healScore > 0.25
 			selector
 			{


slipher:
The `clean non-alerted heal code` changes the semantics in a way that was probably not intended. Wrapping everything in the condition means that the condition will be evaluated each frame. So `action heal` will be aborted once the healScore reaches 0.25. 

The original works differently because `sequence` is stateful and remembers which action it is already doing. (Unlike `selector` which always tries every action from the beginning.)

bmorel:
I've removed it, just in case. No need to delay it further, in any case I foresee a lot of BT rewrites in future 0.53's master, so most of the uglyness will likely be washed away then (replaced by more, likely, considering the size of the changeset I have in mind).

bmorel:
Both cleanup commits have been removed, @necessarily-equal  said LGTM on IRC

PR #120, opened 2022-07-10
make creep sizes 3x horizontal size

bmorel:
Fix https://github.com/Unvanquished/gameplay/issues/18

This makes all alien buildables creep size to 3 times the size of one "horizontal" side size.
I have not tested this yet, I'm notably afraid there will be glitches if I do not obey some undocumented rule about PoT images.
I'm going to run a quick local test for that aspect, but doing the PR so that @necessarily-equal 's systems can catch the PR and host an instance for all players willing to see what it would look like.

### DIFF
--- a/configs/buildables/acid_tube.attr.cfg
+++ b/configs/buildables/acid_tube.attr.cfg
@@ -26,5 +26,5 @@ splashRadius      100
 meansOfDeath      alienBuildable
 
 // Misc
-creepSize         64
+creepSize         120
 dretchAttackable
--- a/configs/buildables/barricade.attr.cfg
+++ b/configs/buildables/barricade.attr.cfg
@@ -25,4 +25,4 @@ splashRadius      100
 meansOfDeath      alienBuildable
 
 // Misc
-creepSize         96
+creepSize         210
--- a/configs/buildables/barricade.attr.cfg
+++ b/configs/buildables/barricade.attr.cfg
@@ -25,4 +25,4 @@ splashRadius      100
 meansOfDeath      alienBuildable
 
 // Misc
-creepSize         210
+creepSize         280
--- a/configs/buildables/booster.attr.cfg
+++ b/configs/buildables/booster.attr.cfg
@@ -26,5 +26,5 @@ splashRadius      100
 meansOfDeath      alienBuildable
 
 // Misc
-creepSize         64
+creepSize         120
 transparentTest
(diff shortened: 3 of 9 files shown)


bmorel:
Ok, so I did a quick test on plat23 map, and:

1) it works
2) it might be too small, but for next release I'd rather have it not increased too much, further actual playing should say if it's good enough or not
3) it feels good, humans are indeed slowed down when walking in alien base even if not kissing the enemy buildings, especially barricade and overmind. This also helps aliens finding a place on which they heal.

bmorel:
I added 2 commits to give additional boost for OM and barricades, because they are intended to be really massive structures.
So, x4 for barricade, to better fulfill role be being annoying for humans, and x5 for OM, to make it clear it's the center of the base.

necessarily-equal:
Testing server at https://play.unvanquished.net/unvanquished.net:27992. Connect with a 0.52.1 client.

bmorel:
To be noticeable, this requires changes in C++, namely, increase the creep effect on humans speed.
What I think can work is this:

```
commit d4529a476c71f2a461c603b1a973bfb02147c14a
Author: Morel Bérenger <berengermorel76@gmail.com>
Date:   Fri Sep 9 17:46:33 2022 +0200

    tweak creep values

diff --git a/src/shared/bg_gameplay.h b/src/shared/bg_gameplay.h
index ee5ff6a80..9107190a4 100644
--- a/src/shared/bg_gameplay.h
+++ b/src/shared/bg_gameplay.h
@@ -116,9 +116,9 @@ extern int   LEVEL4_TRAMPLE_REPEAT;
  */
 
 #define CREEP_BASESIZE          700
-#define CREEP_TIMEOUT           1000
-#define CREEP_MODIFIER          0.5f
-#define CREEP_ARMOUR_MODIFIER   0.75f
+extern const int CREEP_TIMEOUT;
+extern const float CREEP_ARMOUR_MODIFIER;
+extern const float CREEP_MODIFIER;
 #define CREEP_SCALEDOWN_TIME    3000
 
 #define BARRICADE_SHRINKPROP    0.25f
diff --git a/src/shared/bg_misc.cpp b/src/shared/bg_misc.cpp
index 2f5ec3bf6..32bfec62a 100644
--- a/src/shared/bg_misc.cpp
+++ b/src/shared/bg_misc.cpp
@@ -35,6 +35,10 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 // delayed translation - these strings may be passed to Trans_Gettext() later
 #define N_(x) x
 
+const int CREEP_TIMEOUT = 3000;
+const float CREEP_ARMOUR_MODIFIER = 0.62f;
+const float CREEP_MODIFIER = CREEP_ARMOUR_MODIFIER / 2.f;
+
 struct buildableName_t
 {
 	buildable_t number;

```

PR #121, opened 2022-07-11
do not flee if rushScore is high enough

bmorel:
Currently, bots will run away even in the middle of a rush. This is a pretty annoying behavior, as the point of a rush is to deal as much damages as possible.
Since RushScore now returns a value which depend on both the strength of the bot's current configuration and what they could buy in their next life, this should be pretty beneficial, and hopefully will make bots less campy.


necessarily-equal (approved):
LGTM, haven't tested

bmorel:
It is tested by real players since 3 days.

PR #122, opened 2022-07-18
increase most beacons durations

bmorel:
fixes https://github.com/Unvanquished/Unvanquished/issues/1896

It only improves time of 4 beacons, but the other ones seem to be quite entangled in C++, so they shuold probably wait a bit more.

### DIFF
--- a/configs/beacons/attack.beacon.cfg
+++ b/configs/beacons/attack.beacon.cfg
@@ -8,4 +8,4 @@ icon       0 "gfx/feedback/beacons/attack"
 hlIcon     0 "gfx/feedback/beacons/attack"
 inSound    "sound/feedback/beacons/beacon"
 outSound   "sound/feedback/beacons/target-lost"
-decayTime  10000
+decayTime  20000
--- a/configs/beacons/pointer.beacon.cfg
+++ b/configs/beacons/pointer.beacon.cfg
@@ -8,4 +8,4 @@ icon       0 "gfx/feedback/beacons/pointer"
 hlIcon     0 "gfx/feedback/beacons/pointer"
 inSound    "sound/feedback/beacons/beacon"
 outSound   "sound/feedback/beacons/target-lost"
-decayTime  2000
+decayTime  6000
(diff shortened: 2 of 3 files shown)


necessarily-equal:
Sounds good on principle, but to be honest I think the new duration for "defend" is way too long. Especially since the alien base sends those beacons by itself. (And may do so even more if we add some alarm system with https://github.com/Unvanquished/gameplay/issues/13)

bmorel:
> (And the alien base may do so even more if we add some alarm system with Unvanquished/gameplay#13)

This one is still a purely hypothetical feature that we *might* have some day, I do not think it should be considered a strong enough reason to make beacons useful in 0.53.
Maybe the durations are too long with those changes. But I'd rather have them too long, and used, than too short, and ignored. I could even propose a feature for the base to place never-disappearing beacons, and that only when a player (or a bot) disables it explicitly, then the beacon will disappear.
Like, you may have automatic alarms, but they won't shut up automatically.

Also, if that's too long, they we could adjust some values for a point release. It's not hard to do.

necessarily-equal:
Base already send "Defend" beacons, mind you, and it's usually when it's already too late.

bmorel:
Well, no, it's not too late. It means the base is under attack, and that a player should really go there and check out what is happening, or the team might loose precious outpost/base.

But, even if the base beacons are too late, as you say, nothing prevents players to actually use beacons themselves. The reason why they don't is: the duration is *way* too short to be worth loosing time with that circle menu thing.

slipher:
Agree that "defend" duration should not be increased.

30s also seems too long for attack. 20s would be ok.

bmorel:
I reverted the defend beacon change. I have not ignored @slipher 's opinion on attack, I'm just waiting for more talking about it to know which value I should use.

bmorel:
So be it then. If they're still not used because duration is still too small, we can still increase for next version anyway.

necessarily-equal (approved):
LGTM, let's see how that goes

PR #124, opened 2022-07-19
adv granger descriptions says that removing fire brings credits

bmorel:


necessarily-equal (approved):
nice

necessarily-equal:
Maybe terminate the sentence with a dot?

PR #125, opened 2022-07-20
Restore a sane default font size.

DolceTriade:
1.5em is large enough to be readable and the default font size chosen by
RmlUI is too small to be readable.


necessarily-equal:
Replaced by https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/126

PR #129, opened 2022-07-20
human bots: heal even if own medkit

bmorel:
Companion PR for #1934


bmorel:
stupid github was confused. Ignore this.

PR #131, opened 2022-07-21
(almost)automated CSS format

necessarily-equal:


ui/transition_loading.rml
```diff
@@ -66,17 +67,18 @@
...
 
-			statusbox, percentage {
+			statusbox,
+			percentage {
 				color: white;
 				font-size: 1.5em;
 				font-weight: bold;
 				/* Dejavu has monospaced numbers, but Roboto does not
-				   This stops the the percentage number "jiggling" left and right */
+							   This stops the the percentage number "jiggling" left and right */
```
slipher, line 76: The thing messed up the formatting here

slipher (commented):
Reformatting stuff is always controversial because of effectively erasing commit history. If we must do it, I'd rather after the 0.53 merge

ui/transition_loading.rml
necessarily-equal: Fixed

necessarily-equal:
I'd like merging this before 0.53 because it would mean mod have less files to distribute if they are based in master. And less conflicts if anyone starts working on unvanquished-0.53.0 instead of master.

bmorel:
After thinking about it, I think changes to only change code-style are just a problem, and solve nothing. Please link this to me if I ever attempt to do so.
This makes everything more complicated for nothing.
Instead, we should define a script doing the job for us. This is not for humans to do.

necessarily-equal:
I don't think this will ever have to be done twice.

For a start, CSS syntax has the benefit of being much simpler than C++, so it's much less ambiguous which style should be followed. Then, this is mostly about unifying it once to make it easier to read, to avoid wholly unreadable things and unify it once. I don't care too much if it drifts a bit with something slightly different but readable, like someone using `.5em` instead of `0.5em`.

And about using an automatic formatter, well while some css is in its own css files and is easily processable, some is in html files and is thus harder to process automatically

slipher:
> I'd like merging this before 0.53 because it would mean mod have less files to distribute if they are based in master.

Huh?

> And less conflicts if anyone starts working on unvanquished-0.53.0 instead of master.

I've seen at least one RML change go into master recently, so merging the current PR will probably create conflicts. Merging after 0.53 (when there will be only a single existing branch) would avoid conflicts.

DolceTriade:
fwiw, you can ignore commits when doing git blame for history. I don't think rewrite/reformat commits are as breaking as they used to be. However, ideally we do them all in a single/few commits so we can ignore these commits easily.

DolceTriade:
+1 to merging after we release 0.53. Let's move this to 0.54.

necessarily-equal:
Rebased. I hope nothing went wrong when rebasing.

necessarily-equal:
> What the CSS code needs is […] a real cleanup

Well yes, but that would be easier to do if my eyes don't bleed. And the cleanup PR would be easier to read if they don't include formating changes too.

ui/hud_human.rml
```diff
@@ -184,30 +187,27 @@
...
-			unlocked_items
-			{
+			unlocked_items {
 				width: 100%;
 				height: 50%;
 				/* Two-colour system. HACK
-					locked colour = CSS cell-color
-					unlocked colour = CSS color
-				 */
+								locked colour = CSS cell-color
```
slipher, line 201: This spacing seems bad. Otherwise LGTM

DolceTriade (approved):
Thanks for fixing.

ui/hud_human.rml
necessarily-equal: fixed

Issue #128, opened 2022-07-20
Borked spacing in the in-game vote menu


necessarily-equal:
![image](https://user-images.githubusercontent.com/59283660/179999150-73704812-14af-4f9f-8053-aa2673e377f7.png)

PR #132, opened 2022-07-21
Fix spacing around votes

necessarily-equal:
Fixes https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/issues/128


necessarily-equal:
Attempt to fix would be more exact. I have some doubts it works like intended

image

necessarily-equal:
I've added double spacing because it looks nicer IMO: 
image

bmorel:
This breaks the i18n thing.
Spacing is a per-language thing, adding artificial spaces feels like a hack to me.

DolceTriade:
Does this actually break spacing with i18n? It seems like regardless of the language, a littlle more spacing would make things more readable?

This LGTM

ui/menu_ingame.rml
```diff
@@ -33,15 +33,17 @@
 						<h2><translate>Democracy</translate></h2>
 						<indent>
 							<blocklink class="fake">
-								<translate>Global vote:</translate>
+								<!-- note &nbsp; + line break means there are two spaces, but IMO it looks prettier, so well -->
+								<translate>Global vote:&nbsp;</translate>
```
slipher, line 37: Why would you put the `&nbsp;` *inside* the translate tag?

necessarily-equal: whoops, fixed

bmorel:
> Does this actually break spacing with i18n? It seems like regardless of the language, a littlle more spacing would make things more readable?

I guess. It's not very important anyway, and this is a rather trivial change, easy to fix back if problems arise.

PR #137, opened 2022-07-25
hud_bp: Fix BP font size for granger.

DolceTriade:
looks like this now

### DIFF
--- a/ui/hud_bp.rml
+++ b/ui/hud_bp.rml
@@ -8,7 +8,7 @@
 		}
 
 		div {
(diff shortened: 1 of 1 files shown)


DolceTriade:
Ignore pixel corruption in screenshot. It's due to my WM.

necessarily-equal (approved):
Fixes https://github.com/Unvanquished/Unvanquished/issues/1972.

Haven't tested but LGTM.

Note I first thought it was about the bottom left text, which I had noticed but didn't know what to do about it.

PR #138, opened 2022-07-25
Improve circlemenu default state (unselected + unlocked) color and make it more opaque

necessarily-equal:


DolceTriade:
I can't tell any difference with this change

necessarily-equal:
default-style items are a bit darker and more opaque

necessarily-equal:
quote from bmorel:

> Maybe a better solution (than ouline, but outline might also work) would be to have a solid color instead of what looks to be some alpha blending or what-not.

This makes it a (nearly) solid color instead of alpha-blended white. And also it makes it a bit darker so that it's easier to tell the selected/active element

PR #139, opened 2022-07-28
Last-minute UI fixes

slipher:
- Alien BP overlapping class icon
- Revert stamina bar color change, it's too distracting


bmorel:
LGTM

PR #142, opened 2022-07-31
ui/server_browser: Workaround float precision loss: 0% + 90% may be more than 100%

illwieckz:
Workaround float precision loss: 0% + 90% may be more than 100%

- Fix: https://github.com/Unvanquished/Unvanquished/issues/2011

### DIFF
--- a/ui/server_browser.rml
+++ b/ui/server_browser.rml
@@ -10,7 +10,11 @@
 			}
 
 			panel {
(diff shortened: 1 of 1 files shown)


illwieckz:
I tested with 4:3 VGA 640×480, 16:9 HD1080p 1920×1080 and 21:9 UWUHD 3440×1440.

PR #144, opened 2022-08-02
add a possible position indicator under speedometer

bmorel:
This will only show if speedometer is enabled, not sure how to prevent than problem while keeping things reasonably simple. I'm open to suggestions.

### DIFF
--- a/ui/hud_basics.rml
+++ b/ui/hud_basics.rml
@@ -148,13 +148,13 @@
 			font-size: 1.5em;
 		}
 
-		speedometer
+		speedometer,position_indicator
 		{
 			width: 100%;
 			height: 100%;
 		}
 
-		speedometer span
+		speedometer span,position_indicator span
 		{
 			margin: auto auto;
 			padding: 5%;
@@ -229,7 +229,7 @@
 
 	<if cvar="cg_drawSpeed" condition="!=" value="0">
 		<!-- Do we need this if? -->
-		<region class="speedo"><speedometer /></region>
+		<region class="speedo"><speedometer /><position_indicator/> </region>
 	</if>
 
 	<region class="beacon">


ui/hud_basics.rml
```diff
@@ -156,6 +156,18 @@
...
+		{
+			width: 100%;
+			height: 100%;
+		}
+
+		position_indicator span
+		{
+			margin: auto auto;
+			padding: 5%;
+		}
```
necessarily-equal, line 169: I see that this was copy-pasted, how about joining the rules with a "," (e.g. `speedometer, position_indicator {}` and `speedometer span, position_indicator span {}`).

ui/hud_basics.rml
```diff
@@ -225,7 +237,7 @@
 
 	<if cvar="cg_drawSpeed" condition="!=" value="0">
 		<!-- Do we need this if? -->
-		<region class="speedo"><speedometer /></region>
+		<region class="speedo"><speedometer /><position_indicator/> </region>
 	</if>
```
necessarily-equal, line 241: Here I would swap the `region` and the `if`, this way you can use a second if on the second one

necessarily-equal (changes requested):

ui/hud_basics.rml
bmorel: Yes, it was. I know (almost) nothing about CSS so I just go the quick'n dirty way. There might be much cleaner ways to do it.
But, this works, and is pretty useful, as it avoid doing a bind to emit a `/where` command. Examples of use cases are for players to report bugs in a more accurate way than doing a screen shot, or knowing where bots suicide exactly (which requires having bots saying where they suicide, which is a feature that was refused but that I keep on my instances as it can help players identify problems in their base).

ui/hud_basics.rml
bmorel: This makes the code buggy.

ui/hud_basics.rml
bmorel: About the copy-pasting, I'd be ok to not do that, if the code was clean. Otherwise, there's no way I can write clean stuff in that area, since I do not know CSS "subtleties" (to be polite).
One of the requested changes looks like a codestyle nitpick, and the other one makes the code buggy.

bmorel:
One of the requested changes looks like a codestyle nitpick, and the other one makes the code buggy.

ui/hud_basics.rml
```diff
@@ -144,13 +144,13 @@
...
 
-		speedometer
+		speedometer,position_indicator
 		{
 			width: 100%;
 			height: 100%;
 		}
 
-		speedometer span
+		speedometer,position_indicator span
```
necessarily-equal, line 153: This is missing the span:

```suggestion
		speedometer span,
		position_indicator span
```

bmorel: what's the point? I thought initial comment was to reduce duplication, and solution I did works according to tests.

necessarily-equal: The two have a different meaning. `speedometer span` means the `span` inside the speedometer, and `position_indicator` without the span just means the span. While you probably wanted to match the `span`.

To make this more abvious, and to avoid long lines, I slightly prefer to put it on separate lines like I suggested:

```css
speedometer span,
position_indicator span
```

necessarily-equal (approved):

necessarily-equal:
Woooo

PR #145, opened 2022-08-03
Fix /cp font-size scaling on different resolutions

necessarily-equal:
Also make font-size smaller (more reasonable) for warmup message

Fixes https://github.com/Unvanquished/Unvanquished/issues/2013


illwieckz (approved):
LGTM

PR #146, opened 2022-08-03
rmlui: Update assets after the update to master

illwieckz:
The branch is named `ishq/rmlui` despite not being mine because this way it's easier to find:

- https://github.com/Unvanquished/Unvanquished/pull/2027

I spotted that:

```
Warn: Deprecation notice: Element '<progressbar>' renamed to '<progress>', please adjust RML tags accordingly. 
Warn: Deprecation notice: Element '<progressbar>' renamed to '<progress>', please adjust RML tags accordingly. 
```

But that breaks HUD progress bars.

### DIFF
--- a/ui/alien_build.rml
+++ b/ui/alien_build.rml
@@ -115,7 +115,7 @@ tabs {
 		local circlemenu = document:GetElementById("alienBuildMenu")
 		circlemenu.inner_rml = CirclemenuSkeleton(#alien_buildables)
 		for i = 1,#alien_buildables do
-			circlemenu.child_nodes[i].inner_rml = string.format(
+			circlemenu.child_nodes[i+1].inner_rml = string.format(
 				'<button class=%s onclick="AlienBuildableClick(%d,event)" onmouseover="AlienBuildableHover(document,%d)">%s<img src="/%s"/></button>',
 				alien_buildables[i][5], i, i, AvailabilityIcon(alien_buildables[i][5]), alien_buildables[i][1])
 		end
--- a/ui/alien_build.rml
+++ b/ui/alien_build.rml
@@ -14,8 +14,8 @@ body
 .circlemenu icon{
 	display:inline-block;
 	position:absolute;
-	bottom: .8em;
-	right: -.6em;
+	bottom: 0.3em;
+	right: 0.75em;
 	font-size: 1.7em;
 }
 .circlemenu button {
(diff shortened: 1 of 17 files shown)


DolceTriade (approved):

slipher:
This should target something other than master, shouldn't it?

illwieckz:
Ah those are not silenced

```
Warn: Syntax error parsing inline property declaration 'opacity: -nan;'. 
Warn: Breaking change: The 'fill-image' property now needs to be set on the <progress> element, instead of its inner <fill> element. Please update your RCSS source to fix progress bars in this document. 
Warn: Breaking change: The 'fill-image' property now needs to be set on the <progress> element, instead of its inner <fill> element. Please update your RCSS source to fix progress bars in this document. 
Warn: Syntax error parsing inline property declaration 'opacity: -nan;'. 
```

It looks like I fixed the `fill-image` issue but not the `opacity: -nan`.

illwieckz:
> This should target something other than master, shouldn't it?

Yes maybe, we can change the target anyway once we create a better one.

illwieckz:
This may need this on Unvanquished side:

```
diff --git a/src/cgame/rocket/rocket.cpp b/src/cgame/rocket/rocket.cpp
index ace047a83..bff864a39 100644
--- a/src/cgame/rocket/rocket.cpp
+++ b/src/cgame/rocket/rocket.cpp
@@ -394,7 +394,7 @@ void Rocket_Init()
 
        // Add custom client elements
        RegisterElement<SelectableDataGrid>( "datagrid" );
-       RegisterElement<RocketProgressBar>( "progressbar" );
+       RegisterElement<RocketProgressBar>( "progress" );
        RegisterElement<RocketDataSelect>( "dataselect" );
        RegisterElement<RocketConsoleTextElement>( "console_text" );
        RegisterElement<RocketDataSourceSingle>( "datasource_single" );
```

But some progress bar are still full white, and they're always full.

Anyway this is low priority, I just wanted to try to know if that was easy or not.

ui/alien_build.rml
```diff
@@ -125,7 +125,7 @@ tabs {
 </head>
 <body id="alien_build" class="circlemenu alien" onkeydown="CirclemenuHandleKey(event, document, AlienBuildableClick)" onshow="BuildAlienBuildMenu(document)">
 <tabset class="circlemenu">
-<tab></tab>
+<tab>Build</tab>
```
illwieckz, line 128: You may want to do `<tab><translate>Build</translate></tab>` while you're at it, so we can translate the string later.

Same for `<tab>Evolve</tab>` below.

DolceTriade: Done.

illwieckz:
I can't approve because:

> Pull request authors can't approve their own pull requests.[^1]

But it LGTM.

[^1]: That's something GitLab allows anyway.

PR #147, opened 2022-08-04
rework server options UI

bmorel:
This PR:

* merges the bot and server settings dialogs
* adds the server settings dialog to the main menu (still can be improved, notably it's also useful when starting a local game)
* moves server config in the "config" block when in-game (takes less space for nothing)
* add option for g_bot_defaultFill (requires https://github.com/Unvanquished/Unvanquished/pull/1938 )
* add option for g_bot_default_skill

Places the basics for https://github.com/Unvanquished/Unvanquished/issues/2033


bmorel:
Converting this to draft because I'm going to see to merge server-config and bot-config windows in a single one, and this will just be a pain to keep 2 PRs in sync.

bmorel:
The result is rather nice, except that it requires CSS changes that I am very unlikely to be able to do.

bmorel:
Added pages for momentum and build system configurations:

2022-08-04-180458_1362x748_scrot
2022-08-04-180503_1362x748_scrot

It seems that cvars are not yet set when in main menu though, and this alter settings when starting a game, it's annoying and must be fixed before any merge.
Also, it's VERY verbose and repetitive for not much, it would be very nice to find a much better system, not to mention I have no preview of the values. It would be nice to have a proper mechanism like, for example, this: 
2022-08-04-180841_1362x748_scrot instead (if I click on the number, I can even edit it in-place)

bmorel:
fix https://github.com/Unvanquished/Unvanquished/issues/1936

DolceTriade:
Give me a day or so, I'll try to help with this before I leave on vacation

bmorel:
Edited OP because this won't __fix__ things, it will only improve situation. For a complete fix, we'd need cgame to be able to know the default values of cvars.

DolceTriade (approved):
I made a few optional changes. Feel free to accept them.

https://gist.github.com/DolceTriade/153ee8a8d4c9e3cf4c3eb3cdb6f5d8d4

Otherwise, this looks really good. In the future, I'd like to move the cvar descriptions into tooltips to save on real estate on the screen.

DolceTriade:
I made an example gist with tooltip support. Let me know what you think. If you like it we can go ahead and move all the description text to tooltips: https://gist.github.com/DolceTriade/b212d6fa48e7e6d1687068441ebb0131

bmorel:
While I would appreciate *good* tooltips experience, I do not trust our current codebase for that (nor any web-related tech, honestly... even native applications hardly do that correctly...).
Also, the added verbosity in the html files will only make it much more tedious to maintain and improve, and *a lot* of cvars are still missing, that I intend to add in this, more or less slowly depending on my mood and implication.

IMO tooltips and many more changes (grouping of stuff, notably) should be done when there will be a proper way to import the information from [daemon's code](https://github.com/DaemonEngine/Daemon/issues/694).
I consider this PR, and the ones I'll send after, ways to know in which direction those things should do. Working at having a good style, saving space, etc... is just going to be a waste of time until we get most of the cvars put in there. If your motivation is to save space on screen, then imo you're talking about doing premature optimization.

The idea of the tooltip *do* leverage a good point, though: cvars should have both a short and a complete description.

DolceTriade:
I mean what do you define as good?

I think the biggest thing which requires improvement is that there is no value for most of the sliders, which makes it difficult for 
users to understand what they are changing. The reason that there is no value for the sliders is that we have a description of the cvar where the value is. imo, the tooltips are good enough to hold the description so we can display the value next to the slider.

I still do not understand what you mean by <cvar />. If you want to save on typing, we can define a metaelement which does everything we want like

```
<cvarOption cvar="g_warmup" description="Warmup time in seconds where players cannot join a team" format="%.0f" units="seconds" />
```

bmorel:
> I think the biggest thing which requires improvement is that there is no value for most of the sliders, which makes it difficult for users to understand what they are changing. The reason that there is no value for the sliders is that we have a description of the cvar where the value is.

No. The reason I didn't implemented this is that the code to get this is a ridiculously complicated mess.

> I still do not understand what you mean by . If you want to save on typing, we can define a metaelement which does everything we want like `<cvarOption cvar="g_warmup" description="Warmup time in seconds where players cannot join a team" format="%.0f" units="seconds" />`

That would be much better, clearly.
Except that, ideally (but not feasible without engine changes, so we'll have to cope with it for now) data like `description=` should be provided by the C++ side.

DolceTriade:
Yes as do I. That's why I approved it.

PR #148, opened 2022-08-04
Remove oxygen bar

slipher:
This reverts commit ad2896216fc5d6c17852aabea8f5eb2ff1b6916f.

See https://github.com/Unvanquished/Unvanquished/issues/2029#issuecomment-1205645445

### DIFF
--- a/ui/hud_basics.rml
+++ b/ui/hud_basics.rml
@@ -66,18 +66,6 @@
 			width: 50%;
 		}
 
-		region.oxygen {
-			left: 30%;
-			bottom: 10%;
-			width: 40%;
-		}
-		region.oxygen progressbar {
-			height: 1.2em;
-		}
-		region.oxygen progressbar fill {
-			image-color: rgba(108, 200, 255, 200);
-		}
-
 		minimap {
 			display: block;
 			width: 12em;
@@ -235,10 +223,6 @@
 		<beacon_distance /><br />
 		<beacon_age /><br />
 		<beacon_owner />
-	</region>
-
-	<region class="oxygen">
-		<progressbar src="oxygen"/>
 	</region>
 </body>
 </rml>


bmorel:
FWIW I disagree that removing it is a good idea. I'll explain my reasons:

* no official map include enough water to see this
* on ported maps which have enough water, the situation after the fix is an actual improvement, even if, *yes*, it's still buggy

My opinion is that it's better to have, in a beta game, a feature which helps despite bugs that do not seriously impact the official stuff, than a lack of such feature.

slipher:
If I understand correctly, if you spectate one player who is underwater then switch away, you will be stuck with the bar for the rest of the game. (Or if it does go away, it is only a bug that the bar is not drawn when oxygen goes to 0.) This is way too broken to be included in the release.

bmorel:
> If I understand correctly, if you spectate one player who is underwater then switch away, you will be stuck with the bar for the rest of the game

No. If you spectate a player underwater, you won't have the oxybar currently.

slipher:
OK, but if you die underwater then that would happen.

bmorel:
> OK, but if you die underwater then that would happen.

Nothing, since the bar disappear when you run out of air anyway

bmorel:
Oh, actually, when you switch team while underwater, the bar is still around. @necessarily-equal I think you're able to fix that though?

necessarily-equal:
Well the flip side of having this ist hat you would die much less under water 😄

More seriously though, mayne it can break if you get killed by a bullet while you are underwater? have you checked this can actually happen? The progress bar is supposedly protected by a _BOTH condition, but I'm not sure it protects both ppl who are dead and ppl spectating.

bmorel:
> have you checked this can actually happen?

No, because it's hard to test *alone*

bmorel:
what needs to be done is, reset the counter when you're moved to spectator, basically. I'm not familiar with that part of the code, but I'm sure you are.

bmorel:
It have to be pointed out that of the 3 mentioned reasons to remove it:

1. dying while underwater
2. switching followed player while spectating
3. leaving the team

1 is no longer true.

PR #152, opened 2022-09-03
Move server options with other server stuff

necessarily-equal:


ui/menu_ingame.rml
```diff
@@ -92,7 +93,7 @@
 							<!-- <blocklink onclick='Events.pushevent("show demos", event)'><translate>Recorded demos</translate></blocklink> -->
 						</indent>
 
-						<h2><translate>Options</translate></h2>
+						<h2><translate>Client Options</translate></h2>
```
necessarily-equal, line 96: Not sure about this specific change

DolceTriade: I think it's better to leave it as Options or "Game options"

bmorel: Agree, people don't have to know what a client is.

ui/menu_ingame.rml
```diff
@@ -60,8 +60,9 @@
 							<blocklink class="rightfloat" onclick='Events.pushevent("exec disconnect", event)'><translate>Stop server</translate></blocklink>
 							<blocklink class="leftalign" onclick='Events.pushevent("exec pause", event)'><translate>Pause server</translate></blocklink>
 
-							<h2><translate>Map</translate></h2>
+							<h2><translate>Server</translate></h2>
```
bmorel, line 63: I disagree with this change, as imo saying "server" in a local game might be technically valid, it's still confusing. Not sure if it's better or worst than the previous situation with "Map", though.

illwieckz:

PR #153, opened 2022-09-03
Fix server setup menu

slipher:
It doesn't make any sense, but increasing the width of the menu fixed the text flow issue https://github.com/Unvanquished/Unvanquished/issues/2132.

### DIFF
--- a/ui/server_setup.rml
+++ b/ui/server_setup.rml
@@ -11,9 +11,9 @@
 		<!-- GENERAL SERVER CONFIG -->
 		<tab><translate>General</translate></tab><panel>
 			<row>
-				<h3><translate>Free funds frequency</translate></h3>
+				<h3><translate>Free funds period</translate></h3>
 				<input type="range" min="0" max="3600" step="1" cvar="g_freeFundPeriod" />
-				<p><translate>Reward players to stay alive every few seconds</translate></p>
+				<p><translate>Time interval at which players receive free credits/evos</translate></p>
 			</row>
 			<row>
 				<h3><translate>Friendly fire damage multipliers</translate></h3>
@@ -59,8 +59,8 @@
 			</row>
 			<row>
 				<input cvar="g_dretchpunt" type="checkbox" />
-				<h3><translate>Dretch-punt</translate></h3>
-				<p><translate>Dretches can be pounted out-of-the-way by large aliens.</translate></p>
+				<h3><translate>Dretch punt</translate></h3>
+				<p><translate>Dretches can be pushed out of the way by attacking them.</translate></p>
 			</row>
 			<row>
 				<input cvar="g_antiSpawnBlock" type="checkbox" />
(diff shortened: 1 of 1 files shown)


ui/server_setup.rml
```diff
@@ -128,7 +128,7 @@
 			<row>
 				<h3> Half regeneration time </h3>
 				<input type="range" min="0" max="60" step="2" cvar="g_BPRecoveryRateHalfLife"/>
-				<p class="inline"> The duration one will wait before BP recovery gets twice as slow (in minutes) </p>
+				<p class="inline"> The duration one will wait before BP recovery gets twice as slow </p>
```
illwieckz, line 131: Maybe say the max is 1 hour ?

illwieckz (approved):
LGTM.

ui/server_setup.rml
slipher: Well, none of the other ones say their range... though it certainly is a problem that you have no idea what the values are.

bmorel: clearly. In some places, sliders have companion text that display the current value, so this is doable, but I really think a real, dedicated, widget is needed, which would allow to print minimum value, maximum value, current value, and default value. And while at it, to set a specific, exact, value without using the console (with sliders it's very difficult, when  you have too much granularity compared to the size of the slider). This requires more work though, as cgame needs to be able to reach cvars for that.

ui/server_setup.rml
```diff
@@ -3,17 +3,17 @@
...
 		<tabset>
 
 		<!-- GENERAL SERVER CONFIG -->
 		<tab><translate>General</translate></tab><panel>
 			<row>
-				<h3><translate>Free funds frequency</translate></h3>
+				<h3><translate>Free funds period</translate></h3>
 				<input type="range" min="0" max="3600" step="1" cvar="g_freeFundPeriod" />
-				<p><translate>Reward players to stay alive every few seconds</translate></p>
+				<p><translate>Time interval at which players receive free credits/evos</translate></p>
```
bmorel, line 16: The information about "stay alive" which is very important imo lacks, here.

ui/server_setup.rml
```diff
@@ -60,7 +60,7 @@
 			<row>
 				<input cvar="g_dretchpunt" type="checkbox" />
 				<h3><translate>Dretch-punt</translate></h3>
-				<p><translate>Dretches can be pounted out-of-the-way by large aliens.</translate></p>
+				<p><translate>Dretches can be punted out of the way by large aliens.</translate></p>
```
bmorel, line 63: In retrospect, maybe reusing the cvar's word "punt" is not a good idea. This is a rather uncommon english word, and if I didn't enabled it on my server, I might not be able to understand what this means (ok, there's a bit of context, so maybe I would be able, but..).
I think "push" might be easier?

```suggestion
				<p><translate>Dretches can be pushed out of the way by large aliens.</translate></p>
```

bmorel (approved):

ui/server_setup.rml
slipher: Done.

ui/server_setup.rml
slipher: I don't think you need to avoid dying, just to keep playing.

bmorel: After checking: it's every 120s of being in a "spawned" state, that is: alive. When one dies, the time spent in "spectator" state is not considered for the "reward".
I'm not too sure how to phrase that though (and I really think this is a very camper-friendly feature, too... but it's needed for when someone joins a game after few minutes :/)

slipher: The way I see it the alive thing is mostly irrelevant unless you are out of spawns and losing anyway. Mentioning alive time just gives the false impression that you need to avoid getting killed to get them.

bmorel: True.

necessarily-equal:
Sounds good, haven't tested

illwieckz:
I merge this so it can be shipped with 0.53.2.

PR #154, opened 2022-09-05
Adjust speedometer and position indicator sizes and font

necessarily-equal:
The intent is to make it less distracting

### DIFF
--- a/ui/hud_basics.rml
+++ b/ui/hud_basics.rml
@@ -145,6 +145,9 @@
 		position_indicator {
 			width: 100%;
 			height: 100%;
+			/* Dejavu has monospaced numbers, but Roboto does not
+			   This stops the numbers "jiggling" left and right */
+			font-family: Dejavu Sans;
 		}
 
 		speedometer span,


ui/hud_basics.rml
```diff
@@ -145,6 +145,9 @@
 		position_indicator {
 			width: 100%;
 			height: 100%;
+			/* Dejavu has monospaced numbers, but Roboto does not
+			   This stops the numbers "jiggling" left and right */
+			font-family: Dejavu Sans;
```
DolceTriade, line 150: Do we ship this font? iirc, we use unifont for a true monospaced font.

necessarily-equal: Yes we do. We use this trick elsewhere too, I didn't invent it. For example the clock is using DejaVu Sans too.

bmorel:
I have doubts about the font size reduction. The text is not that big... I have not seen the result of those changes, though (why can't people just post screenshots for UI changes?)

necessarily-equal:
I've reverted the size change. I did change the size before the font, and I think DejaVu Sans has small characters

ui/hud_basics.rml
necessarily-equal: Also note that only the numbers are constant-width

necessarily-equal: And it is embedded inside unvanquished.dpk
> Loaded font face 'DejaVu Sans' [regular] from 'fonts/DejaVuSansCondensed.ttf'.

DolceTriade:
Pls post screenshot

necessarily-equal:
looks like this 
image

DolceTriade (approved):

PR #155, opened 2022-09-06
add an outline to all text to vastly improve readability

cu-kai:
Adds 2px outlines to `cp` messages, and 1px outlines to everything else for improved visibility across the game.

# NOTE:
### This is more noticeable at lower resolutions, but should have the desired effect for everyone. For the below screenshots, I am playing at 1800 x 1000.

A reference has been provided at the bottom of this post showing this change on 1024x768.

Examples:
===

### Chat text:
image

### Map/game information:
image

### Messagemode and tutorial:
image

### "cp" message:
image

### "cp" message with colour:
image

### "cp" message with colour similar to skybox (this is really where on vanilla, the contrast sucks):
image

### 1024 x 768
image

### DIFF
--- a/ui/hud_overlay.rml
+++ b/ui/hud_overlay.rml
@@ -6,6 +6,7 @@
 			width: 100%;
 			height: 100%;
 			font-size: 18dp;
+			font-effect: outline( 2px #111 );
 		}
 		div {
 			position: absolute;
--- a/ui/shared/basics.rcss
+++ b/ui/shared/basics.rcss
@@ -10,6 +10,7 @@ body {
 	color: #999999;
 	cursor: ui/assets/cursor;
 	font-size: 12dp;
+	font-effect: outline( 1px #111 );
 }
 
 p {


necessarily-equal:
Thanks. Will test later. I'm wondering though, should we apply this only for HUD and text draw over the 1st person game view (and not main menu and in-game menu and options)?

cu-kai:
I personally don't mind how it looks in the menu (you can test this by joining bunker now, and opening the menu). But that's not really a decision for me to make.

DolceTriade (approved):

slipher:
It would be good to have a screenshot of menus.



> This scales by your screen resolution.

Well, if you're using `px` units, then it shouldn't scale. `dp` would be a unit that scales with resolution.

cu-kai:
> It would be good to have a screenshot of menus.
> 
> > This scales by your screen resolution.
> 
> Well, if you're using `px` units, then it shouldn't scale. `dp` would be a unit that scales with resolution.

Sorry yes, I think I meant the perception of it scales with resolution. It indeed does not scale with screen resolution, but you notice it more at lower resolutions. That is not a bad thing IMO.

cu-kai:
Here's a screenshot of the menus, as requested:

unvanquished_2022-09-07_113041_000

DolceTriade:
I definitely think this is a huge improvement. I was a little jarred initially, but once I played with it more, I really appreciated it. Would definitely prefer to merge this first and improve later.

slipher:
> On another note, I think I've found where the readability problem I'm feeling is coming from: it's when there is a black border around some blue (^4) text or most other cases where the color is close enough to the color of the black outline. Maybe to improve on that we can use white border around these. I think this can be improved on later though.

Maybe we could have Rocket_QuakeToRML add `class="darktext"` or something like that to spans using a dark foreground color. Then CSS targeting this class could be added.

necessarily-equal:
Thanks, merging

PR #156, opened 2022-09-10
server_setup: Replace info text with tooltips

DolceTriade:
Looks like

### DIFF
--- a/ui/server_setup.rml
+++ b/ui/server_setup.rml
@@ -1,6 +1,7 @@
 <rml>
 	<head>
 		<link type="text/rcss" href="/ui/shared/basics.rcss" />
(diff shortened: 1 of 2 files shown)


necessarily-equal:
I'm somewhat conflicted, because it works well even if it is a bit hidden. I fear it might not be discoverable enough, or a bit cumbersome to read for someone who discovers the menu for the first time.

But yeah, from a technical point of view it works.

DolceTriade:
The biggest flaw with the current menu is that it doesn't print the current value because it has the descriptions so you have these sliders but no basis for them. That in itself is a big problem.

DolceTriade:
Freem LGTM'd on IRC

bmorel (approved):

PR #158, opened 2022-09-17
buildables: Add new icons made by @Gireen

DolceTriade:
Humans:

Aliens:


bmorel:
Few non-blocking remarks:

1) the rocket-pod could use something to make the difference from the mgtturet more obvious, like a rocket or explosion "badge" somewhere
2) medistation and telenode look very alike. I would remove the "rings" of the medistation, replacing them with more health cross, to make semantic contrast higher
3) it would probably better if we could have the source of the png files, as I doubt the png were edited "as is". If this is not possible because source is lost, then it's sad, but otherwise would be preferable.

This is still overall much better than current situation, so I'm very in favor for merging this.

illwieckz:
One concern I have is that they aren't centered in menu circles.

As bmorel noticed, telenode and medipad looks good if looked at separately but can be easily confused. What if the faded lines are kept on medipad but removed on telenode, with a stylized human standing over the medipad instead?

illwieckz:
@Gireen do you have SVG for them?

Gireen:
I should still have the SVG somewhere but i'll have to look into that, those icons are already a few years old. 

The cross form the medipad could be a bit bigger and the lines a bit more transparent then would be the difference to telenode stronger.

Gireen:
I found the source and modified it a bit
icons

I have all icons in a single SVG for easier editing as a whole. If there are no other changes needed then i would split them to separate files and make a PR for it.

DolceTriade:
That would be awesome @Gireen !

bmorel:
What is the 1st icon? I can't see what it is at all... I thought might be drill, then reactor, but neither.
Otherwise, I think the medistat/telenode problem is fixed, now, but I wonder if you could get a human icon on top of telenode icon to make it even more obvious.
Answer might be yes or no, it's not important enough to justify keeping our current icons though. I prefer those.

DolceTriade:
I think they look fine as they are. The drill and RC look sufficiently different.

DolceTriade:
Closing in favor of https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/163

PR #160, opened 2022-09-20
icons: new armour and bsuit icons

illwieckz:
larmour, marmour and bsuit icons from https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/99 by @Bob-Vador.

I also optimized the png (40% saved size) and stripped some dust from svg editors tends to write (like position and size of Inkscape window on Bob-Vador screen… 🤦‍♀️️).


illwieckz:
I noticed Inkscape has some options to write simple & optimized svg, here it saves 31% of the svg file size so I experimented with it a bit. Maybe we can recommend using that before writing to a git repo.

Gireen:
Is it really necessary to optimize images for every commit? Could that part be moved to the packaging process?

illwieckz:
Optimizing a release package (with release formats) and not committing files using source format but with useless size that will live forever in git repository are two different things. Every time we edit an image and save it in png in the repository, the git repo will keep the variant forever, so if we can reduce 30~40% size of every committed blob file this isn't a bad thing.

illwieckz:
By looking at the svg source I noticed the bsuit icon wasn't really white, so I fixed the svg and reexported the png.

bmorel:
Then maybe we should avoid having the png files in the repo, when we have source svg

illwieckz:
People already don't use urcheon when urcheon provides similar build features for the other formats (the special case of svg-to-image isn't implemented though). Also for now this specific package is meant to be usable without relying on urcheon.

Even if we had an svg-to-image build we may not want to commit the position and the size of the inkscape window on the screen of the latest contributor (and other metadata about his own inkscape preferences and even sometime metadata about his own file system layout) every time we edit the svg anyway.

DolceTriade:
Also keep in mind we need to support different tools: daemon/unv aren't the only tools that need to read these files. NetRadiant for example may want to leverage icons and such from these packs and it can't load svgs or crn

illwieckz:
> Also keep in mind we need to support different tools: daemon/unv aren't the only tools that need to read these files. NetRadiant for example may want to leverage icons and such from these packs and it can't load svgs or crn

Those special icons aren't read by NetRadiant and now only the DEPS file from this repository is needed by NetRadiant, but the comment is true in general as it may be a concern for other dpkdirs.

NetRadiant is meant to also be able to read assets from source repositories, and we can create some textures meant to be displayed in game that have svg sources, like logos and other symbols. So if we implement svg-to-image in urcheon, we would have to add this to the `prepare` urcheon command. This `urcheon prepare` command is already used for converting `.iqe` to `.iqm` files in source tree, and producing `.shader` files with sloth in source tree, this would have to be extended to also take care of svg. I see nothing that cannot be done, but that's a lot of work and that would require to add yet another dependency to urcheon : an svg-to-png converter.

To be 100% compatible with inkscape svg features, an svg exporter may have to be inkscape itself… That's the same reason why I haven't implemented an xcf exporter in urcheon because I was afraid of adding gimp as dependency of urcheon…

Note: NetRadiant now knows how to read crn, in fact you're the one who implemented it if I'm right 😁️ (all our asset formats NetRadiant may have to read from dpk are now supported).

DolceTriade (approved):

bmorel:
> People already don't use urcheon when urcheon provides similar build features for the other formats

I'm certain you know some of the reasons why. Not sure if you actually went around checking with people themselves the reasons, though.
Anyway, this is kind of out-topic here, such a discussion belongs to Urcheon projects (and I'm not really willing to spend my time at it on github).

Here, the question is about merging those icons into upstream, considering the fact that, this time, there's source.
To me, this should be merged: LGTM.

bmorel (approved):

PR #161, opened 2022-09-26
Print level info on map loading screen

illwieckz:
The `<levelinfo/>` element can contain the map name or more complex strings built up by the game with other data, like a translatable string `<mapname> by <author>`.

Requires this to have data to print:

-  https://github.com/Unvanquished/Unvanquished/pull/2226

### DIFF
--- a/ui/shared/basics.rcss
+++ b/ui/shared/basics.rcss
@@ -13,6 +13,10 @@ body {
 	font-effect: outline( 1px #111 );
 }
 
+div {
+	display: block;
+}
+
 p {
 	display: block;
 	margin-bottom: 0.3em;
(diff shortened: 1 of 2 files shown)


slipher:
Is the long name actually the best thing to show to the user? For map selection, voting, etc. the short name is what you need to know...

Don't use nonexistent element types like `serverlevelname`. It's strange that this does not cause a warning. (Yes I see that someone already that in this file but it's bad). Instead you can put the styles directly on `levelname`, or if that doesn't work for some reason, make the enclosing element `<div class="serverlevelname">` and put styles for `.serverlevelname`.

If it looks like the screenshot on the linked PR, I think it would be good to have more vertical space between the level info and the server name.

PR #162, opened 2022-09-29
UI for callvote fillbots

sweet235:
We already have the `kickbots` vote in the menu. This gives users a menu entry to vote for `fillbots`.

### DIFF
--- a/ui/callvote_fillbots.rml
+++ b/ui/callvote_fillbots.rml
@@ -1,0 +1,26 @@
+<rml>
+	<head>
+		<link type="text/rcss" href="/ui/shared/basics.rcss" />
+		<link type="text/template" href="/ui/shared/window.rml" />
+	</head>
+	<body id="callvote_fillbots" template="window" style="width: 35em; margin: 1em;">
+		<h1><translate>Fill with bots</translate></h1>
+		<br />
+		<p>How many players (users and bots) should teams have?</p>
+		<form onsubmit='Events.pushevent("execForm \"callvote $action$ $amount$\"; hide", event)'>
+			<row>
+				<select name="action">
+					<option value="fillbots">both teams</option>
+					<option value="fillbots_aliens">aliens</option>
+					<option value="fillbots_humans">humans</option>
+				</select>
+				<h3><translate>Team</translate></h3>
+			</row>
+			<row>
+				<input type="text" class="text" style="width: 12em;" name="amount"/>
+				<h3><translate>Amount</translate></h3>
+			</row>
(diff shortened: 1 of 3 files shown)


slipher:
Screenshot?

We don't have anything more specific than a free text field for choosing the number?

slipher:
> We don't have anything more specific than a free text field for choosing the number?

Looking at https://mikke89.github.io/RmlUiDoc/pages/rml/forms.html I guess not, assuming a slider is too annoying.

sweet235:
Screenshot:
a

> We don't have anything more specific than a free text field for choosing the number?

I was thinking about a slider or a dropdown menu, but then found it hard to get the upper limit from the server (that limit is in a cvar).

sweet235:
If a player enters an invalid number like `100` (or any invalid string), the server will respond with:
```
callvote: number must be non-negative and smaller than 11
```
The exact value of `11` depends on `g_maxVoteFillBots`. That is how current master behaves.

So I think that on the second try, a user will know what they may safely enter.

ui/callvote_fillbots.rml
```diff
@@ -0,0 +1,26 @@
...
+		<link type="text/template" href="/ui/shared/window.rml" />
+	</head>
+	<body id="callvote_fillbots" template="window" style="width: 35em; margin: 1em;">
+		<h1><translate>Fill with bots</translate></h1>
+		<br />
+		<p>How many players (users and bots) should teams have?</p>
+		<form onsubmit='Events.pushevent("execForm \"callvote $action$ $amount$\"; hide", event)'>
+			<row>
+				<select name="action">
+					<option value="fillbots">both teams</option>
```
slipher, line 13: Option names could be capitalized and translated

Otherwise LGTM

sweet235: Will do

sweet235:
One more question: entering a value of 0 does the same as the (now redundant) kickbots vote. Additionally, I am experimenting with a vote to change bot skill on the fly. Do you think we should make a single window for all bot votes? The menu has grown quite long already.

slipher:
> One more question: entering a value of 0 does the same as the (now redundant) kickbots vote.

Technically it is not always exactly the same since kickbots can kick bots that were not added by bot fill.

> Additionally, I am experimenting with a vote to change bot skill on the fly. Do you think we should make a single window for all bot votes? The menu has grown quite long already.

I don't know; it might be confusing that there are several options but you can only change one of them at a time.

sweet235:
Then let's do it like this, at least for now.

DolceTriade (approved):

PR #166, opened 2022-12-12
Throw grenades instead of just dropping them on the floor

necessarily-equal:
This also lowers the added speed a bit to avoid throwing them unreasonably far

### DIFF
--- a/configs/missiles/firebomb.attr.cfg
+++ b/configs/missiles/firebomb.attr.cfg
@@ -10,8 +10,8 @@ clipmask           MASK_SHOT
 size               3
 
 trajectory         TR_GRAVITY
-speed              400
-lag                0.0 // TODO: Add a fitting lag
+speed              320
+lag                0.8 // portion of the velocity of the player which is added to the projectile speed
 
 doKnockback
 bounceHalf
--- a/configs/missiles/grenade.attr.cfg
+++ b/configs/missiles/grenade.attr.cfg
@@ -10,8 +10,8 @@ clipmask           MASK_SHOT
 size               3
 
 trajectory         TR_GRAVITY
-speed              400
-lag                0.0 // TODO: Add a fitting lag
+speed              320
+lag                0.8 // portion of the velocity of the player which is added to the projectile speed
 
 doKnockback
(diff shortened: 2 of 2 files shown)


slipher:
So the formula is like `missile->speed + missile->lag * (player speed)`? So it seems the patch decreases the human's throwing strength a bit, but allows him to do it while running using his momentum :thinking:

necessarily-equal:
Exactly

cu-kai:
I'm not entirely convinced this is a problem which needs solving, especially as it has not been tested in gameplay yet. It feels like a solution in search of a problem.

sweet235:
Newton would be pleased. I, however, am worried that a human using a jetpack can reach speeds that make a grenade fly through half of the map. Let's test this.

sweet235:
Let me elaborate on that. Grenades are terribly strong. A common question from new players is why they are so expensive. The answer is, of course, that they can wreak havoc in the alien base. Accidentally throwing a grenade into the human base regularly requires admin intervention. They are among the strongest weapons in the human arsenal.

Grenades, as they are now, require a human to be very close their point of denotation when throwing them; at least in non-vertical situations. We should be very careful with changing this.

Here is a somewhat related video, for fun: https://www.youtube.com/watch?v=j_DIR--yRK0

necessarily-equal:
For the record, I don't intend on merging this in the current state 
since 3 people exprimed strong doubts about the idea. I guess this 
would need more testing.

illwieckz:
> Throw grenades instead of just dropping them on the floor

I'm curious about this sentence, as grenade are already thrown. You can, for example, look high over a barricade to throw the grenade over the barricade to reach the alien base behind it. One can reproduce this in plat23 successfully.

Aiming with grenade has its own learning curve but it's really a useful game mechanics and it works for me™.

Example: https://dl.illwieckz.net/b/unvanquished/shots/grenade-throwing/20230117-160835-001.unvanquished-grenade-throwing.webm

illwieckz:

necessarily-equal:
Let's close it for now because players were not enthusiastic about it.

PR #167, opened 2022-12-15
camper.bt: roam freely if it does not work near reactor/overmind

sweet235:
Current camper.bt makes bots using it freeze when their main buildable disappears. This makes them roam instead.

### DIFF
--- a/bots/camper.bt
+++ b/bots/camper.bt
@@ -18,11 +18,7 @@ selector
 				action heal
 			}
 
-			fallback
-			{
-				action roamInRadius( E_H_REACTOR, 500 )
-				action roam
-			}
+			action roamInRadius( E_H_REACTOR, 500 )
 		}
 	}
 
@@ -38,11 +34,9 @@ selector
 			{
 				action evolve
 			}
-			fallback
-			{
-				action roamInRadius( E_A_OVERMIND, 500 )
-				action roam
-			}
+			action roamInRadius( E_A_OVERMIND, 500 )
 		}
 	}
+
+	action roam
 }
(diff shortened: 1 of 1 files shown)


bots/camper.bt
```diff
@@ -17,8 +17,12 @@ selector
 			{
 				action heal
 			}
-			
-			action roamInRadius( E_H_REACTOR, 500 )
+
+			fallback
```
slipher, line 21: The `fallback` is unnecessary. You would only need that if you really want it to keep going to the target selected by `roam` even after a main structure was constructed. But immediately returning to the main structure seems better. So just add one line `action roam` after the roamInRadius.

necessarily-equal: Sounds reasonable here

bots/camper.bt
```diff
@@ -34,7 +38,11 @@ selector
...
 				action evolve
 			}
-			action roamInRadius( E_A_OVERMIND, 500 )
+			fallback
+			{
+				action roamInRadius( E_A_OVERMIND, 500 )
+				action roam
+			}
 		}
 	}
```
necessarily-equal, line 47: By the way, I think you could even just add the action roam at the very end

```suggestion
	}
	action roam
```

sweet235: Did that, works.

sweet235:
I consider this to be more of an example file than an actual config. But in any case, users expect it to work somehow. Should we add roaming near own spawns too if the main buildable cannot be found?

necessarily-equal:
> > Should we add roaming near own spawns too if the main buildable 
> cannot be found?
> 

Oh yeah good idea

sweet235:
Let me try this first. I'm worried that eggs on ceilings might confuse the bots.

slipher:
LGTM

PR #168, opened 2022-12-21
Fix camper bot fighting

necessarily-equal:
There is a subtule difference between the two structures. Notably, with this new `sequence` format, the action fight continues if the enemy is no longer visible. The enemy is thought to be "not visible" more often than you think.

Suggested testing protocol:
* add camper human bots
* sit on the armory as a dragoon or anything else. You should be surprisingly safe without this change.

### DIFF
--- a/bots/camper.bt
+++ b/bots/camper.bt
@@ -8,7 +8,6 @@ selector
 		action fight
 	}
 
-	
 	condition team == TEAM_HUMANS
 	{
 		selector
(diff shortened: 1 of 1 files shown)


slipher:
Why is there a separate commit to remove the empty line you added in the first commit?

necessarily-equal:
That's because I made this PR on my phone, and github doesn't allow editing commits there. This PR can simply be squashed

Le 21 décembre 2022 12:25:50 GMT+01:00, slipher ***@***.***> a écrit :

sweet235:
I can confirm that this fixes the (unreported) issue.

slipher:
LGTM

PR #169, opened 2023-01-02
presets: set archive when modifying cvar, also drop unimplemented MSAA, fix Unvanquished/Unvanquished#2340

illwieckz:
Set archive when modifying cvar

- fix Unvanquished/Unvanquished#2340

Also reset `r_ext_multisample` which isn't implemented yet, since multisampling isn't implemented yet it means the values were never tested and then, if we set or keep a value set to this cvar, people may have the bad surprise of a potential performance hit the day they upgrade the engine to a version that implements the feature.

### DIFF
--- a/presets/graphics/high.cfg
+++ b/presets/graphics/high.cfg
@@ -11,7 +11,6 @@ seta r_heatHaze 1
 seta r_bloom 1
 seta r_subdivisions 4
 seta r_imageMaxDimension 2048
-seta r_ext_multisample 8
 seta r_ext_texture_filter_anisotropic 8
 seta r_textureMode "GL_LINEAR_MIPMAP_LINEAR"
 seta r_clear 1
--- a/presets/graphics/high.cfg
+++ b/presets/graphics/high.cfg
@@ -1,24 +1,24 @@
-r_dynamicLight 2
-r_halfLambertLighting 1
-r_rimLighting 1
-r_vertexLighting 0
-r_normalMapping 1
-r_deluxeMapping 1
-r_specularMapping 1
-r_physicalMapping 1
-r_reliefMapping 0
-r_heatHaze 1
-r_bloom 1
-r_subdivisions 4
-r_imageMaxDimension 2048
-r_ext_multisample 8
-r_ext_texture_filter_anisotropic 8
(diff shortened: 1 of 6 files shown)


illwieckz:
For `r_ext_multisample` we may prefer to entirely remove it from presets and add a code in Dæmon to reset it, code we would remove the day we implement multisampling.

slipher:
LGTM

illwieckz:
See also this PR to reset `r_ext_multisample` on engine side:

- https://github.com/DaemonEngine/Daemon/pull/773

Anyway, doing it on engine side doesn't prevent us to also do it on preset side.

presets/graphics/high.cfg
```diff
@@ -11,7 +11,7 @@ seta r_heatHaze 1
 seta r_bloom 1
 seta r_subdivisions 4
 seta r_imageMaxDimension 2048
-seta r_ext_multisample 8
+reset r_ext_multisample
```
slipher, line 14: Now you can drop this

illwieckz: I just done it.

illwieckz:
I added a commit to also remove the MSAA option from the UI.

ui/options_graphics.rml
```diff
@@ -76,22 +76,9 @@
...
-						<option value="8">8×</option>
-						<option value="16">16×</option>
-					</select>
-					<h3><translate>Multisampling level (MSAA)</translate></h3>
-					<p><translate>Anti-aliasing smoothes out “jaggy edges“ but is very costly to your frame-rate.</translate></p>
-				</row>
-
 				<row>
 					<input cvar="r_FXAA" type="checkbox" />
 					<h3><translate>Fast approximate anti-aliasing (FXAA)</translate></h3>
```
necessarily-equal, line 81: Is this one implemented?

necessarily-equal (approved):
One question, LGTM because it improves things

ui/options_graphics.rml
illwieckz: Yes, it doesn't work as expected but it is implemented:

- https://github.com/DaemonEngine/Daemon/issues/533

slipher:
LGTM

PR #170, opened 2023-01-16
ui: rework the map loading screen

illwieckz:
This makes the information (map name, author…) more readable whatever the level shot is. I also set the loading bar to the Unvanquished cyan color, also make sure the progress bar text has a bit of bottom margin.

Some graphics designer may want to do something better, but at least this improves readability without being so bad.

Before:

[map loading screen](https://dl.illwieckz.net/b/unvanquished/wip/map-loading-screen/20230116-170923-000.unvanquished.jpg)

After:

[map loading screen](https://dl.illwieckz.net/b/unvanquished/wip/map-loading-screen/20230116-170958-000.unvanquished.jpg)

Before:

[map loading screen](https://dl.illwieckz.net/b/unvanquished/wip/map-loading-screen/20230116-170940-000.unvanquished.jpg)

After:

[map loading screen](https://dl.illwieckz.net/b/unvanquished/wip/map-loading-screen/20230116-171009-000.unvanquished.jpg)

### DIFF
--- a/ui/transition_loading.rml
+++ b/ui/transition_loading.rml
@@ -28,7 +28,7 @@
 			barbox {
 				display: block;
 				width: 100%;
-				height: 2em;
+				height: 2.4em;
 
 				position: absolute;
 				left: 0em;
@@ -43,7 +43,7 @@
 			}
 
 			progress {
-				image-color: #999999cc;
+				image-color: #1ab0b5;
 			}
 
 			statusbox {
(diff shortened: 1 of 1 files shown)


ui/server_browser.rml
```diff
@@ -22,6 +22,7 @@
 				height: 30em;
 				width: 100%;
 				display: block;
+				overflow: hidden;
```
slipher, line 25: This shouldn't be here

slipher (commented):
Looks good

ui/server_browser.rml
illwieckz: Oh, that's a stray one from #171…

PR #172, opened 2023-01-24
add option to enable/disable light styles in presets and options

illwieckz:
`r_lightStyles` is a recent cvar added in https://github.com/DaemonEngine/Daemon/pull/751

Light Styles are precomputed dynamic lights for static map objects: the map compiler bake multiple versions of the same lightmap with different parameters (for example: light on, light off) and the renderer blends them in game to create some effects. It's much cheaper than realtime dynamic lights and not that heavy, Tremulous maps already used this trick a lot.

- Add option to enable disable light styles in graphics option in game,
- Set values in graphics preset, since it's considered “not that heavy”, only the “lowest” profile disables this for now,
- Also reword the dynamic light option description while I was there.

### DIFF
--- a/presets/graphics/high.cfg
+++ b/presets/graphics/high.cfg
@@ -1,3 +1,4 @@
+seta r_lightStyles 1
 seta r_dynamicLight 2
 seta r_halfLambertLighting 1
 seta r_rimLighting 1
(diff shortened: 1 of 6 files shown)


illwieckz:
Before releasing 0.54, we may want to rename `r_lightStyles` to `r_lightStyle` without plural form to be consistent with other cvar names like `r_dynamicLight` that use singular form.

illwieckz:
There are many other cvars that use plural forms:

```
r_vboModels
r_softShadows
r_bloomPasses
```

There is just no consistency.

Let's merge this as is.

slipher:
`lightStyles` is better anyway because `lightStyle` sounds like it should be a choice between different types of lighting.

PR #175, opened 2023-03-15
no longer use "action heal" in BT

bmorel:
This PR simply removes all calls to `action heal` as the action itself is not needed: it's really just a matter of changing goal and going there.

### DIFF
--- a/bots/default_aliens.bt
+++ b/bots/default_aliens.bt
@@ -24,14 +24,7 @@ selector
 					{
 						condition ( healScore > 0.5 && percentHealth( E_GOAL ) > 0.3  && random > 0.3 )
 						condition baseRushScore < 0.75
-						selector
-						{
-							action changeGoal( E_A_BOOSTER )
-							action changeGoal( E_A_OVERMIND )
-							action changeGoal( E_A_SPAWN )
-						}
-						action resetStuckTime
-						action moveToGoal
+						behavior heal
 					}
 				}
 			}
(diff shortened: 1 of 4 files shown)


bmorel:
I have tested it works with both aliens and humans.

slipher:
Looking at the action heal implementation, some of the logic we would lose:
- Failing when the medistation is being used by another player
- Using a larger radius for the booster

bmorel:
> Failing when the medistation is being used by another player

yes, I think for this I'd need a better version of goalChange, one that would take some filtering options.

> Using a larger radius for the booster

This is a valid problem for all alien building, what is needed is a way to check if the bot is within range of it's target. This utility would actually be useful for future changes too.

There is another benefit of this change, outside of the C++ code reduction though: big aliens won't consider themselves stuck when they're just waiting for heal.
Also, in a (hopefully) near future, I want to have actions to rotate the agent (to search for enemy), instead of just idling in the heal spot. Using a split "action heal" will make this much more easier.

sweet235:
I noticed that toggling a rush by setting the goal directly has its drawbacks. It can make bots run into the enemy defenses blindly, and die there. On my server, I currently make the bots roam near these buildings instead of setting them as goals. That makes the bots choose their own targets.

bmorel:
Weird, because in both cases they should be alerted by enemy before continuing the rush/flee.

illwieckz:

This PR should be redone on https://github.com/Unvanquished/Unvanquished instead.

I let this PR open so the one redoing the PR on the other repo can close it when it's redone, so we know what PR is redone or not yet.

bmorel:
tbh I'm not sure it's such a good idea in current code status.
The fact this submodule has finally be killed will make this easier to have something actually finished and more reliable in future.
I'm not sure it's worth the effort to migrate that.

PR #176, opened 2023-03-16
add defend.bt, follow.bt, stay-here.bt

sweet235:
This adds three example behavior files for bots. They are reasonably well tested. My server currently uses nearly the same files, and Der Bunker server currently uses exactly these files.

What these files do:

- `follow.bt` makes a bot follow another player/bot. To make bot 1 follow player 2, use `/bot behavior 1 follow 2`. To make a bot on your team follow you while playing, use `/tactic follow 1`.
- `stay-here.bt` makes a bot stay near a specified point. To make bot 1 stay near point `(100, 200, 300)`, use `/bot behavior 1 stay-here 100 200 300`. To make a bot on your team stay where you currently are, use `/tactic stay-here 1`.
- `defend.bt` is a more elaborate version of `camper.bt`, handling things like human bots repairing or preventing damaging the own base.

There is some duplication. I am aware of that. But as these are config files, I think it is desirable to make them stand-alone.

### DIFF
--- a/bots/defend.bt
+++ b/bots/defend.bt
@@ -1,0 +1,187 @@
+selector
+{
+	behavior unstick
+
+	condition ( team == TEAM_ALIENS )
+	{
+		selector
+		{
+			condition ( aliveTime > 1500 && healScore < 0.5 )
+			{
+				action evolve
+			}
+
+			sequence
+			{
+				condition alertedToEnemy
+				selector
+				{
+					decorator timer( 1000 )
+					{
+						selector
+						{
+							sequence
+							{
+								condition alertedToEnemy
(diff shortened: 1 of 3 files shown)


cu-kai (approved):
These are used daily on @sweet235's server and also during any games taking place on Bunker. No problems to report, this is well tested.

sweet235:
I should add that these files only differ from `default.bt` when a bot is not alerted to an enemy. On very small maps, bots are almost always alerted to enemies.

There is now `g_bot_alienSenseRange`, reducing this number will reduce the likelyhood of bots being alerted to enemies. Currently, this even affects human bots to some degree. (Don't ask me why)

bmorel:
with that much real life testing, I wonder if a review is really useful. This is a clear improvment, and one of the reasons why sweet's server is that popular.

LGTM

bmorel (approved):

PR #178, opened 2023-03-27
remove `gfx/colors/black` as it looks unused

illwieckz:
remove `gfx/colors/black` as it looks unused

We have plenty of other black textures like the engine builtin `$blackImage` for internal use and `textures/shared_colors/black` in `tex-common` dpkdir for use in assets (maps, models…).


bmorel:
If it's not used, then let's remove it. lgtm

PR #180, opened 2023-04-09
Limit bot fill slider to safe values

slipher:
9 is the maximum bots you can have per team without running out of slots given a solo player and the default sv_maxclients of 20.

Fixes https://github.com/Unvanquished/Unvanquished/issues/1812.

### DIFF
--- a/ui/server_setup.rml
+++ b/ui/server_setup.rml
@@ -192,7 +192,7 @@
 					<h3><translate>Minimum bots per team</translate></h3>
 					<tooltip><translate>Fill teams until they have enough players</translate></tooltip>
 				</parent>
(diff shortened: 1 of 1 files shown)


bmorel:
LGTM.
I think we'd need a way to control number of bots in each team, but it's out of scope and I don't think anybody is planning to work on GUI those days, so let's get this fixed.

necessarily-equal (approved):
LGTM

illwieckz:

This PR should be redone on https://github.com/Unvanquished/Unvanquished instead.

I let this PR open so the one redoing the PR on the other repo can close it when it's redone, so we know what PR is redone or not yet.

PR #182, opened 2023-04-11
ui: remove whitespaces from translatable strings that are not meant to be translated

illwieckz:
Remove whitespaces from translatable strings that are not meant to be translated.

Fixes:

```
Error in file ./ui/help_gameplay.rml on line 59: translated string should not contain newlines
Error in file ./ui/help_gameplay.rml on line 131: translated string should not contain newlines
2 markup error(s) found
Error in file ./ui/welcome.rml on line 24: leading or trailing whitespace in translated string
Error in file ./ui/welcome.rml on line 24: translated string should not contain newlines
2 markup error(s) found
```


slipher:
Duplicate of #181

illwieckz:
Oh right.

PR #184, opened 2023-04-12
Decommissioning the submodule

illwieckz:
This repository is now merged back into Unvanquished repository as a subfolder. This repository should not be used anymore.

See https://github.com/Unvanquished/Unvanquished/pull/2625

The `DEPS` file is kept to keep compatibility with old NetRadiant that would look for this package to know the default packages to load.

Newer NetRadiant now looks for the `res-leveleditor` package to know the default packages to load.


necessarily-equal (approved):
F

